PR #24, opened 2024-12-20
feat: mpmc channels

tisonkun:


tisonkun:
kanal's and flume's implementation seem good enough.

PR #41, opened 2025-03-06
fix: semaphore wake and drop race condition

tisonkun:
fyi @andylokandy @leiysky

### DIFF
--- a/mea/src/internal/waitlist.rs
+++ b/mea/src/internal/waitlist.rs
@@ -112,6 +112,8 @@ impl<T> WaitList<T> {
             let next = self.nodes[idx].next;
             self.nodes[prev].next = next;
             self.nodes[next].prev = prev;
+            self.nodes[idx].prev = idx;
+            self.nodes[idx].next = idx;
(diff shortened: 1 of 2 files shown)


mea/src/semaphore/tests.rs
```diff
@@ -159,6 +159,27 @@ fn acquire_then_drop() {
     assert_eq!(s.available_permits(), 1);
 }
 
+#[test]
+fn wake_then_drop() {
```
tisonkun, line 163: Always panic with "invalid key" without the fix above.

mea/src/internal/waitlist.rs
```diff
@@ -112,6 +112,8 @@ impl<T> WaitList<T> {
             let next = self.nodes[idx].next;
             self.nodes[prev].next = next;
             self.nodes[next].prev = prev;
+            self.nodes[idx].prev = idx;
+            self.nodes[idx].next = idx;
```
tisonkun, line 116: Now, when a node is removed, we set the pointer to itself so that we don't hold dangling pointers.

PR #51, opened 2025-05-30
feat: oneshot channel

tisonkun:
This is derived from https://github.com/faern/oneshot/.

Special thanks to @faern and @Cassy343 for the original work. I made my own for mea, since mea needs not support synchronized receiving functions. And this can simplify the code a lot.

1. `Receiver::drop` no longer needs to handle RECEIVING or AWAKING state. Because only `try_recv` avaialble and any RECEIVING state would only happen on the `Recv` future.
2. Reuse the same busy loop logic for `Recv` when handling AWAKING state in a poll.

This also make use of the patch https://github.com/faern/oneshot/pull/61 and I suppose that it provides a better impl for implementing `Recv` as a future, rather than under the name `AsyncReceiver` and thus no `async_channel`.

Other refs -

* https://github.com/faern/oneshot/pull/17
* https://github.com/faern/oneshot/pull/20
* https://github.com/faern/oneshot/pull/23
* https://github.com/faern/oneshot/pull/60

### DIFF
--- a/mea/src/oneshot/mod.rs
+++ b/mea/src/oneshot/mod.rs
@@ -77,7 +77,6 @@ use std::fmt;
 use std::future::Future;
 use std::future::IntoFuture;
 use std::hint;
-use std::marker::PhantomData;
 use std::mem;
 use std::mem::MaybeUninit;
 use std::pin::Pin;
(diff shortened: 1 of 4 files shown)


mea/src/oneshot/mod.rs
tisonkun, line 1: Already invariant due to `UnsafeCell<T>`.

PR #52, opened 2025-05-31
feat: impl mpsc + fork atomicbox

tisonkun:
The performance does not even regress .. Actually, flume uses a similar channel state lock at the very beginning.

### DIFF
--- a/mea/src/mpsc/mod.rs
+++ b/mea/src/mpsc/mod.rs
@@ -1,0 +1,71 @@
+//!
+
+use crate::internal::Mutex;
+use std::collections::VecDeque;
+use std::future::poll_fn;
+use std::sync::Arc;
+use std::task::{Context, Poll, Waker};
+
+#[cfg(test)]
+mod tests;
+
+///
+pub fn unbounded<T>() -> (UnboundedSender<T>, UnboundedReceiver<T>) {
+    let state = Arc::new(Mutex::new(UnboundedState {
+        buffer: VecDeque::new(),
+        rx_task: None,
+    }));
+
+    (
+        UnboundedSender {
+            state: state.clone(),
+        },
+        UnboundedReceiver {
+            state: state.clone(),
+        },
+    )
+}
+
+///
+#[derive(Clone)]
+pub struct UnboundedSender<T> {
+    state: Arc<Mutex<UnboundedState<T>>>,
+}
+
+impl<T> UnboundedSender<T> {
+    ///
+    pub fn send(&self, value: T) {
+        let mut state = self.state.lock();
+        state.buffer.push_back(value);
+        if let Some(waker) = state.rx_task.take() {
+            waker.wake();
+        }
+    }
+}
+
+///
+pub struct UnboundedReceiver<T> {
+    state: Arc<Mutex<UnboundedState<T>>>,
+}
+
+impl<T> UnboundedReceiver<T> {
+    ///
+    pub async fn recv(&mut self) -> Option<T> {
+        poll_fn(|cx| self.poll_recv(cx)).await
+    }
+
+    fn poll_recv(&mut self, cx: &mut Context<'_>) -> Poll<Option<T>> {
+        let mut state = self.state.lock();
+        if let Some(value) = state.buffer.pop_front() {
+            Poll::Ready(Some(value))
+        } else {
+            state.rx_task = Some(cx.waker().clone());
+            Poll::Pending
+        }
+    }
+}
+
+struct UnboundedState<T> {
+    buffer: VecDeque<T>,
+    rx_task: Option<Waker>,
+}
--- a/mea/src/mpsc/mod.rs
+++ b/mea/src/mpsc/mod.rs
@@ -1,71 +1,285 @@
-//!
+// Copyright 2024 tison <wander4096@gmail.com>
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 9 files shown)


mea/src/mpsc/mod.rs
```diff
@@ -0,0 +1,310 @@
...
+    fn drop(&mut self) {
+        drop(self.sender.take()); // drop the sender; this closes the channel if this is the last sender
+
+        match self.state.senders.fetch_sub(1, Ordering::AcqRel) {
+            1 => {
+                // If this is the last sender, we need to wake up the receiver
+                // so it can observe the disconnected state.
+                let waker = self.state.rx_task.swap(null_mut(), Ordering::AcqRel);
+                if !waker.is_null() {
+                    unsafe { (*waker).wake_by_ref() }
```
SkiFire13: This will leak both the `Waker` and the `Box` it was stored in.

mea/src/mpsc/mod.rs
```diff
@@ -0,0 +1,310 @@
...
+    /// If the receiver has been dropped, this function returns an error. The error includes
+    /// the value passed to `send`.
+    pub fn send(&self, value: T) -> Result<(), SendError<T>> {
+        // SAFETY: The sender is guaranteed to be non-null before dropped.
+        let sender = self.sender.as_ref().unwrap();
+        sender.send(value).map_err(|err| SendError(err.0))?;
+
+        let waker = self.state.rx_task.swap(null_mut(), Ordering::AcqRel);
+        if !waker.is_null() {
+            unsafe { (*waker).wake_by_ref() }
```
SkiFire13: Same here.

mea/src/mpsc/mod.rs
```diff
@@ -0,0 +1,310 @@
...
+        poll_fn(|cx| self.poll_recv(cx)).await
+    }
+
+    fn poll_recv(&mut self, cx: &mut Context<'_>) -> Poll<Option<T>> {
+        match self.receiver.try_recv() {
+            Ok(v) => Poll::Ready(Some(v)),
+            Err(std::sync::mpsc::TryRecvError::Disconnected) => Poll::Ready(None),
+            Err(std::sync::mpsc::TryRecvError::Empty) => {
+                let waker = Box::into_raw(Box::new(cx.waker().clone()));
+                self.state.rx_task.store(waker, Ordering::Release);
```
SkiFire13: If `poll_recv` was previously called, or the `Future` returned by `recv` was dropped, and no `send` has been performed since then, then `rx_task` will contain the `Waker` from that previous invocation and this `store` will leak both the `Waker` and the `Box` it was stored in.

SkiFire13 (commented):
Overall looks good but has a few leaking issues. My suggestion would be to write (or use) an abstraction like [`atomicbox::AtomicOptionBox`](https://docs.rs/atomicbox/latest/atomicbox/struct.AtomicOptionBox.html) that ensures you never leak (or double-free) the waker.

mea/src/mpsc/mod.rs
tisonkun: Thanks for pointing this out!

These two should be able to fix by replacing with `unsafe { Box::from_raw(waker).wake() }`.

tisonkun:
> [atomicbox::AtomicOptionBox](https://docs.rs/atomicbox/latest/atomicbox/struct.AtomicOptionBox.html)

@SkiFire13 Thanks for sharing this utility. I'll take a look. Previously I'd use `AtomicWaker` but there are some comments (https://github.com/rust-lang/futures-rs/issues/2955) indicates it may be dropped from tokio/future-channel eventually.

mea/src/mpsc/mod.rs
tisonkun: Yeah.

Not sure if add a drop glue within `UnboundedReceiver`'s Drop impl would help. Let me think ..

tisonkun: Use `AtomicOptionBox` should be better. Updating ...

SkiFire13: > Not sure if add a drop glue within UnboundedReceiver's Drop impl would help

No, since you can call `poll_recv` or `recv` multiple times without dropping `UnboundedReceiver`. Adding drop glue to the return value of `recv` could help, but it would solve only half the issue, so I wouldn't recommend it.

---

The way I would fix it would be to use `swap` instead of `store`, and drop the `Waker` and `Box` that was stored in `rx_task` (if not null).

SkiFire13:
> Previously I'd use AtomicWaker but there are some comments (https://github.com/rust-lang/futures-rs/issues/2955) indicates it may be dropped from tokio/future-channel eventually.

If I understand the comments correctly it may be dropped from futures-rs (not sure where you saw tokio/future-channel, that doesn't seem to exist) due to being replaced with the `atomic-waker` crate.

mea/src/mpsc/mod.rs
tisonkun: Thanks for your advice. I'd adopt `AtomicOptionBox` so that it handles all resource drops properly.

> That crate seems better than allocating a Box though, so I would go for it.

Not quite. When you put it in the implementation of the mpsc channel, you need to always share the data between threads, which requires allocation anyway.

While AtomicWaker requires one less indirection to use, it increases the atomic ops needed to lock the state properly, while AtomicOptionBox uses one swap in every path. I think the latter is more clearly correct and easier to follow.

Issue #54, opened 2025-05-31
Implement mpsc BoundedSender and BoundedReceiver


tisonkun:
It should be as easy as add a buffer-sized `Semaphore` to the channel state.

1. Factor out the current unbounded impl to a new file, said `src/mpsc/unbounded.rs`
2. Add the new implementation under `src/mpsc/bounded.rs` as described above.
3. Add some tests for the bounded API. You may take a look at how tokio, async-channel and futures-channel use a mpsc channel and their tests.

CookiePieWw:
I'd like to take this :) I'd like to checkout the implementation of `Semaphore` and `mpsc::BoundedSender` first. Since I'm a very beginner of async rust, would you like to suggest other materials?

tisonkun:
Async Rust is somehow a different language to Sync Rust ..

There is no general material to talk about what you need to know. The official [async book](https://rust-lang.github.io/async-book/) tells only a very specific perspective. You may directly play with the primitives, read some related RFC like:

* https://rust-lang.github.io/rfcs/3668-async-closures.html
* https://rust-lang.github.io/rfcs/2394-async_await.html
* https://rust-lang.github.io/rfcs/3513-gen-blocks.html

The key point here is to understand how Rust compiles `.await` and how `poll` works.

tisonkun:
@CookiePieWw You can get started with step 1 to get familiar with the codebase.

tisonkun:
Also, `BoundedSender` can have a `try_send` method and the corresponding `TrySendError`.

PR #55, opened 2025-06-03
refactor: move unbounded mpsc to a single file

CookiePieWw:
As title. Part of #54 

I walk through the implementations (with copilot definitely :). I'd like to assume the async block is compiled as a state machine with every await as a stop point. And there is a runtime consistently poll all state machines, make it possible for these state machine to continue if possible.

The unbounded mpsc seems quite easy since there's only one await point, so we can explain it like:
```C++
enum State {
    Pending,
    Finished
}

auto poll() {
    switch (self.state) {
        Pending:
            if (self.has_sth()) {
                self.state = Finished;
                return something;
            } else {
                return nothing;
            }
    }
}

int main() {
    for future in all_futures {
        if future.state == State::Finished {
            remove(future);
        } else {
            poll(future);
        }
    }
}
```

It's a rough model though. Seems the true async world is much more complicated.

### DIFF
--- a/mea/src/mpsc/error.rs
+++ b/mea/src/mpsc/error.rs
@@ -32,7 +32,7 @@ impl<T> SendError<T> {
     pub fn into_inner(self) -> T {
         self.0
     }
-    
+
     /// Creates a new `SendError` with the given message.
     pub(super) fn new(msg: T) -> SendError<T> {
         SendError(msg)


mea/src/mpsc/mod.rs
```diff
@@ -14,300 +14,8 @@
...
-    fn fmt(&self, fmt: &mut fmt::Formatter<'_>) -> fmt::Result {
-        match *self {
-            TryRecvError::Empty => "receiving on an empty channel".fmt(fmt),
-            TryRecvError::Disconnected => "receiving on a closed channel".fmt(fmt),
-        }
-    }
-}
-
-impl std::error::Error for TryRecvError {}
+pub use unbounded::{unbounded, SendError, TryRecvError, UnboundedReceiver, UnboundedSender};
```
tisonkun: I suppose that `SendError` and `TryRecvError` are shared between bounded and unbounded impls, so either keep it in mod.rs or factor out to a new error.rs file should be fine.

CookiePieWw: Updated with an `error.rs` and some fmt fix. Since we cannot construct a `SendError` with private field, I made the field `pub(crate)`.

tisonkun: I push a commit to use a `pub(super) fn new` constructor instead. Expose field can expose extra access.

PR #56, opened 2025-06-06
feat: add bounded mpsc

CookiePieWw:
As title.

A problem is that we use `self.state.tx_permits.release(u32::MAX as usize);` when droppint the `BoundedReceiver`, would this cause overflow when available permits are large?

### DIFF
--- a/mea/src/mpsc/bounded.rs
+++ b/mea/src/mpsc/bounded.rs
@@ -141,13 +141,9 @@ impl<T> BoundedSender<T> {
                     let poll = pin!(&mut self.acquire).poll(cx);
 
                     value = match self.sender.try_send(value) {
-                        Ok(()) => {
-                            self.sender.state.tx_permits.release_if_nonempty(1);
-                            return Poll::Ready(Ok(()));
-                        }
+                        Ok(()) => return Poll::Ready(Ok(())),
                         Err(TrySendError::Disconnected(value)) => {
-                            self.sender.state.tx_permits.notify_all();
-                            return Poll::Ready(Err(SendError::new(value)));
+                            return Poll::Ready(Err(SendError::new(value)))
                         }
                         Err(TrySendError::Full(value)) => value,
                     };
--- a/mea/src/mpsc/bounded.rs
+++ b/mea/src/mpsc/bounded.rs
@@ -17,6 +17,8 @@
 
 use std::fmt;
 use std::future::poll_fn;
+use std::future::Future;
+use std::pin::pin;
 use std::sync::atomic::AtomicUsize;
 use std::sync::atomic::Ordering;
 use std::sync::Arc;
@@ -25,6 +27,7 @@ use std::task::Poll;
 use std::task::Waker;
 
 use crate::atomicbox::AtomicOptionBox;
+use crate::internal::Acquire;
 use crate::internal::Semaphore;
 use crate::mpsc::error::TrySendError;
 use crate::mpsc::SendError;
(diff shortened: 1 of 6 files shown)


CookiePieWw:
@tisonkun Thanks for your suggestion :) I tried to use `notifiy_all` of internal semaphore to replace the `release(u32::MAX as usize)`, but `notifiy_all` only wakes all waiters without adding permits. The problem is that since it does not add any permits, following calls to `try_recv` will still return an error indicating the channel is full, instead of notifying the caller that the channels are disconnected. Thus I add a `is_close` flag and checks it before acquiring permits to fix that.

Hope this finds you well. There's no rush at all from my side though :)

CookiePieWw:
> Could you make or refer to a test case where following calls to try_recv will still return an error indicating the channel is full?

Sry there's a typo, `try_recv` shoud have been `try_send`. The case is like:
```rust
#[tokio::test]
async fn main() {
    let (tx, rx) = mpsc::bounded::<i32>(1);
    tx.try_send(1).unwrap(); // the channel is full
    assert_eq!(tx.try_send(2), Err(TrySendError::Full(2)));
    drop(rx); // it's fine. senders above don't need any permits.
    assert_eq!(tx.try_send(3), Err(TrySendError::Disconnected(3))); // panic here, since following calls to semaphore.acquire finds there's no permits available, and returns full before checking if the channel has disconnected.
}
```

[This branch](https://github.com/CookiePieWw/mea/tree/test-fail) add the case and reference impl using `notify_all` only without `is_close` flag, the test case [here](https://github.com/CookiePieWw/mea/blob/14328d5ef88355c90464e8859478b5a3a9d0483c/mea/src/mpsc/tests.rs#L258) (same as above) failed.

CookiePieWw:
@tisonkun Ping again since I'm not sure if the email/notification is sent. The comment above is edited to add the ping instead of a direct ping. Sry if bothered, it's not in a hurry.

The notification mechanism of github is quite ambigous for me. Last time I edited my comment multiple times in a issue under an apache project and I found all versions in the mail list...

tisonkun:
Yeah. I was notified. I'm considering the design and corner case implement details, perhaps I'll back here in one week or two.

tisonkun:
@CookiePieWw I push a new commit for using std's bounded mpsc to simplify code.

You can review https://github.com/cratesland/mea/pull/56/commits/f456cefd6995b853c9eaf7422497815dceba91c0 for double check.

The performance seems improved by about 40% (bounded pressure tests 500ms to 300ms)

mea/src/mpsc/bounded.rs
```diff
@@ -0,0 +1,328 @@
...
+            Ok(()) => return Ok(()),
+            Err(TrySendError::Disconnected(value)) => return Err(SendError::new(value)),
+            Err(TrySendError::Full(value)) => value,
+        };
+
+        loop {
+            self.state.tx_permits.acquire(1).await;
+            value = match self.try_send(value) {
+                Ok(()) => {
+                    self.state.tx_permits.release_if_nonempty(1);
```
tisonkun: Maybe this line is unnecessary, but it's not a bug - at most we notify_all and senders may find they failed to send (due to full).

@CookiePieWw If you have time and interest, you can try to reasoning the new impl to see if we can safely drop this line. Or is there any bug.

mea/src/mpsc/bounded.rs
```diff
@@ -0,0 +1,328 @@
...
+                    return Ok(());
+                }
+                Err(TrySendError::Disconnected(value)) => {
+                    self.state.tx_permits.notify_all();
+                    return Err(SendError::new(value));
+                }
+                Err(TrySendError::Full(value)) => value,
```
tisonkun: Possible case:

* suspicious wake up
* a `try_send` comes first

mea/src/mpsc/bounded.rs
CookiePieWw: Sure, I'll have a try :)

mea/src/mpsc/bounded.rs
```diff
@@ -0,0 +1,328 @@
...
+    pub async fn send(&self, value: T) -> Result<(), SendError<T>> {
+        let mut value = match self.try_send(value) {
+            Ok(()) => return Ok(()),
+            Err(TrySendError::Disconnected(value)) => return Err(SendError::new(value)),
+            Err(TrySendError::Full(value)) => value,
+        };
+
+        loop {
+            self.state.tx_permits.acquire(1).await;
+            value = match self.try_send(value) {
```
tisonkun: Possible race - before acquire, try_recv call `release_if_nonempty` and find it's empty, then the next acquire would never be wake up.

tisonkun: Try fix with manually poll - efcb2ca5d82cf32ef594776c755b68fcb9c198b8

tisonkun:
@CookiePieWw I'd merge this patch now. Before releasing it, you can go through the related document to see if we miss something.

You can search the unbounded mpsc docs and see if anywhere we miss a bounded mpsc channel's doc entry.

mea/src/mpsc/bounded.rs
```diff
@@ -0,0 +1,358 @@
...
+
+/// Creates a bounded mpsc channel for communicating between asynchronous
+/// tasks with backpressure.
+///
+/// A `send` on this channel will wait if the buffer of the channel is full until a
+/// `recv` is called on the receiver, which will consume the message and
+/// free up space in the buffer.
+#[track_caller]
+pub fn bounded<T>(buffer: usize) -> (BoundedSender<T>, BoundedReceiver<T>) {
+    assert!(buffer > 0, "mpsc bounded channel requires buffer > 0");
```
tisonkun, line 44: Another follow-up idea is to support rendezvous channel (buffer = 0).

The underneath implementation would be greatly different, while the interface can remain the same.

To keep code maintenable, if we implement a rendezvous channel, I'd keep them in a separate `mod`.

PR #60, opened 2025-07-20
feat: reduce atomicbox argument

tisonkun:
cc @andylokandy 

Hide the argument first. I'd check the real constraints here later.

### DIFF
--- a/mea/src/atomicbox/atomic_box.rs
+++ b/mea/src/atomicbox/atomic_box.rs
@@ -60,27 +60,18 @@ impl<T> AtomicBox<T> {
     /// any values.  `other` is moved into `self`; the value previously in
     /// `self` is returned.
     ///
-    /// `order` must be either `Ordering::AcqRel` or `Ordering::SeqCst`,
-    /// as other values would not be safe if `T` contains any data.
-    ///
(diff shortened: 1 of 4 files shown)


mea/src/atomicbox/atomic_box.rs
```diff
@@ -60,88 +60,56 @@ impl<T> AtomicBox<T> {
...
-    pub fn swap_mut(&self, other: &mut Box<T>, order: Ordering) {
-        match order {
-            Ordering::AcqRel | Ordering::SeqCst => {}
-            _ => panic!("invalid ordering for atomic swap"),
-        }
-
+    pub fn swap_mut(&self, other: &mut Box<T>) {
         let other_ptr = Box::into_raw(unsafe { ptr::read(other) });
-        let ptr = self.ptr.swap(other_ptr, order);
+        let ptr = self.ptr.swap(other_ptr, Ordering::AcqRel);
```
andylokandy, line 112: can be Relaxed

tisonkun: As described above, I'd modify the API first, and change the implementation later with more verification.

PR #61, opened 2025-07-26
feat: implement MappedMutexGuard

### LINKED ISSUE #59: Implement MappedMutexGuard
See https://github.com/cratesland/mea/pull/58 as an example.

Also reference to:

* https://docs.rs/tokio/latest/tokio/sync/struct.MappedMutexGuard.html
* https://doc.rust-lang.org/std/sync/struct.MappedMutexGuard.html
* https://github.com/rust-lang/libs-team/issues/260
* https://github.com/rust-lang/rust/issues/117108

We can implement for a matrix of:

* Owned/(NonOwned)
* Lock/ReadLock/WriteLock

in total six variants.

orthur2:
First and foremost, thank you for assigning this issue to me. Here is the work for review.

This PR implements the six variants of mapped guards for both `Mutex` and `RwLock`, as proposed in issue #59:

- `MappedMutexGuard` (NonOwned, Mutex)
- `OwnedMappedMutexGuard` (Owned, Mutex)
- `MappedRwLockReadGuard` (NonOwned, RWLockRead)
- `OwnedMappedRwLockReadGuard` (Owned, RWLockRead)
- `MappedRwLockWriteGuard` (NonOwned, RWLockWrite)
- `OwnedMappedRwLockWriteGuard` (Owned, RWLockWrite)
- and also docs and tests

The difficult parts of the programming process were the strenuous derivation of the `Send` and `Sync` traits, as well as managing lifetimes.

The current implementation may have some issues. Please feel free to review the code thoroughly and point out any problems or even the smallest errors. I would be very grateful for your feedback.

It's worth noting that I hesitated for a long time on whether to use `try_map()` or `filter_map()`, even though I saw that @tisonkun's PR #58 used `try_map()`. In my view, the name has a significant impact. While `filter_map()` is a well-regarded naming choice, and in fact, many designs in the Rust community have evolved from `try_map()` to `filter_map()`, I ultimately chose to use `try_map`. My reasoning is that from a user's perspective, the most critical piece of information is that this function might fail and they need to handle its `Result` return value. The name `filter_map` focuses more on the function's internal mechanism, whereas `try_map` better communicates the operation's external behavior.

Another naming challenge I faced was with the "owned" guards. Although I initially listed them as `OwnedMappedMutexGuard`, `OwnedMappedRwLockWriteGuard`, and `OwnedMappedRwLockReadGuard` when I took the issue. However, during development, I was torn between `Owned(MappedMutexGuard)` and `Mapped(OwnedMutexGuard)`. If we follow a functional approach of `Operation(Object)`, then `Mapped(OwnedMutexGuard)` would be the name that best reflects the process. However, I ultimately chose `OwnedMappedMutexGuard` because I feel it better describes the properties of the Mapped....Guard `struct` itself. (Owned or NonOwned)

If you have different ideas, please share your views.

**What I should mention that I do else:**

I moved the `read()`, `try_read()`, `write()`, `try_write()`, `read_owned()`, `write_owned()`, `try_read_owned()`, and `try_write_owned()` methods of the `RwLock<T>` struct from their individual files (`read_guard.rs`, `write_guard.rs`, etc.) into `mod.rs`. I believe it's better to have the method implementations (`impl TheType { ... }`) in the same file as the type definition (`struct TheType { ... }`). All these methods belong with the `RwLock` struct definition.

I fixed a documentation build warning in the `mpsc` module (`unresolved link to UnboundedSender::send`,unresolved link to `BoundedSender::send`, etc.). This was a minor, unrelated warning, but with my preference for keeping things tidy, I fixed it here. I apologize for forgeting creating a separate PR for this, as it is technically out of scope for this issue. Since it was a small change, hope you don't mind me adding it here.

**Some related work recommendations for the future:**

- **Write Lock Downgrading:** I considered adding functionality to downgrade a write lock to a read lock when implementing `MappedRwLockWriteGuard`. However, I noticed the existing `WriteGuard` didn't have a similar function, so I omitted it. This would be better suited for a separate feature request. After a quick search, I found that both the Rust standard library and Tokio provide `downgrade()` and even `try_downgrade_map`.
- **Mutex Module Refactoring:** It might be a good idea to split the `mutex/mod.rs` file into separate files like `mutexguard.rs`, `ownedmutexguard.rs`, etc., similar to the structure of the `rwlock` module. I was originally going to do this while working on the issue but decided against it for this PR to avoid extensive code changes that would complicate the review process.

Closes #59


mea/src/mutex/mod.rs
```diff
@@ -47,7 +47,7 @@
 //! }
 //!
 //! let final_value = mutex.lock().await;
-//! assert_eq!(*final_value, 3); // 0 + 1 + 2
+//! assert_eq!(*final_value, 3); // 0 + 0 + 1 + 2
```
tisonkun: Unrelated change?

mea/src/mpsc/error.rs
```diff
@@ -16,8 +16,8 @@ use std::fmt;
 
 /// An error returned when trying to send on a closed channel.
 ///
-/// Returned from [`UnboundedSender::send`] and [`BoundedSender::send`] if the corresponding
-/// [`UnboundedReceiver`] has already been dropped.
+/// Returned from [`crate::mpsc::UnboundedSender::send`] and [`crate::mpsc::BoundedSender::send`] if the corresponding
+/// [`crate::mpsc::UnboundedReceiver`] has already been dropped.
 ///
 /// The message that could not be sent can be retrieved again with
 /// [`SendError::into_inner`].
```
tisonkun: I'd prefer to `use crate::mpsc::UnboundedSender` above or use ``[`UnboundedSender::send`]: crate::mpsc::UnboundedSender`` in doc comment. Since your diff changes the final displayed docs.

mea/src/mutex/mod.rs
```diff
@@ -168,7 +176,7 @@ impl<T: ?Sized> Mutex<T> {
     /// use mea::mutex::Mutex;
     ///
     /// let mutex = Mutex::new(1);
-    /// let mut guard = mutex.try_lock().expect("mutex is locked");
+    /// let mut guard = mutex.try_lock().expect("mutex should be available");
```
tisonkun: Unrelated changes? Ditto others.

Looks like something vibe coding touched. Please try to avoid this kind of change unless you have a good reason.

mea/src/mutex/mod.rs
```diff
@@ -314,6 +324,125 @@ impl<T: ?Sized> DerefMut for MutexGuard<'_, T> {
...
+    ///         db.users.get_mut(&admin_id)
+    ///     } else {
+    ///         None
+    ///     }
+    /// }).expect("admin user should exist");
+    ///
+    /// assert_eq!(&*admin_email_guard, "admin@example.com");
+    /// # }
+    /// ```
+    pub fn try_map<U, F>(mut orig: Self, f: F) -> Result<MappedMutexGuard<'a, U>, Self>
```
tisonkun: Although I share the same understanding with your rationale in PR description, I tend to use `filter_map` to provide users a familiar experience from sync mutexes.

mea/src/rwlock/read_guard.rs
```diff
@@ -13,94 +13,19 @@
...
+use crate::rwlock::{RwLock, MappedRwLockReadGuard};
 /// RAII structure used to release the shared read access of a lock when dropped.
 ///
 /// This structure is created by the [`RwLock::read`] method.
+///
+/// See the [module level documentation](crate::rwlock) for more.
 #[must_use = "if unused the RwLock will immediately unlock"]
 pub struct RwLockReadGuard<'a, T: ?Sized> {
-    lock: &'a RwLock<T>,
+    pub(super) lock: &'a RwLock<T>,
```
tisonkun: I keep `impl<T: ?Sized> RwLock<T> { pub async fn read { .. } }` inside this file to avoid `pub(super)` since unfortunately Rust cannot express pub(super) for readonly.

tisonkun:
> Write Lock Downgrading
> Mutex Module Refactoring

Both sound reasonable. I just don't have time to implement time before :D

mea/src/mpsc/error.rs
orthur2: Fine.

mea/src/mutex/mod.rs
orthur2: Thanks for the detailed feedback and for catching that. Yeah, I used LLM to help with docs and tests and accidentally accepted this unrelated suggestion. I'm sorry for my carelessness. I agree it's out of scope and will revert it to keep this PR focused.

mea/src/mutex/mod.rs
orthur2: Good point. The name `filter_map` is indeed more consistent with the APIs of `std::sync::Mutex` and `Tokio::sync::Mutex`, which lowers the learning curve for users. I will make the change.

mea/src/rwlock/read_guard.rs
orthur2: Oh, I see. I hadn't thought of that; it's definitely a worthy trade-off. It might feel inconsistent that if only put the `read` and `try_read` functions in `[read_guard.rs](http://read_guard.rs/)`, so I'll probably roll back this move change.

tisonkun:
Thanks for your contribution!

> Write Lock Downgrading
> Mutex Module Refactoring

If you're interested in these, welcome to drop new patches. And you can ping me as a reviewer.

orthur2:
> LGTM. I push a new commit to updates some coments. Rest is great.

Thanks a lot for the review and for handling the comments.Glad I could contribute!

orthur2:
> Thanks for your contribution!
> 
> > Write Lock Downgrading
> > Mutex Module Refactoring
> 
> If you're interested in these, welcome to drop new patches. And you can ping me as a reviewer.

Well, I would like to continue contributing when I have some free time.

Issue #63, opened 2025-08-08
Bug: Memory leak in map/filter_map for Owned* guards


orthur2:
**Summary**

When revisiting the implementation while preparing to add `downgrade()`, I found `Arc` strong-reference leaks in `map`/`filter_map` of three owned mapped guards. The bug was introduced by using `ManuallyDrop` together with `Arc::clone()`, which leaves the original strong reference stranded in `ManuallyDrop` and never dropped. This bug was caused by my attempt to prevent the original Guard from releasing its lock and to preserve the `Arc` reference that provides the lock for the `MappedGuard`. In fact, I should have transferred ownership of the `Arc` instead of sharing it.

 I am very sorry for writing incorrect Unsafe Rust code. I will fix this issue.

**Suggested Fix**

I will submit a Pull Request that replaces guard.lock.clone() with unsafe { std::ptr::read(&guard.lock) } in all affected map and filter_map methods. The PR will also include the new tests.

Issue #62, opened 2025-08-02
Implement Write Lock Downgrading


orthur2:
Currently, if a task holding a write lock finishes modifying data and wants to allow concurrent reads, it must drop the write lock and then re-acquire a read lock. This process is inefficient and can introduce potential race conditions in the time between dropping the write lock and acquiring the read lock.

This was identified as a potential future improvement during the work on PR #61.

This issue proposes adding a `downgrade()` method to `RwLockWriteGuard`. This would allow a task to atomically convert an exclusive write lock into a read lock (`RwLockReadGuard`).This improves performance and throughput by allowing read operations to proceed immediately without contention from other pending write locks.

This is a common and proven feature in other major Rust libraries:

- **Rust Standard Library**: [`RwLockWriteGuard::downgrade()`](https://doc.rust-lang.org/std/sync/struct.RwLockWriteGuard.html#method.downgrade)
- **Tokio**: [`RwLockWriteGuard::downgrade()`](https://docs.rs/tokio/latest/tokio/sync/struct.RwLockWriteGuard.html#method.downgrade)

And I would like to work on this issue.

orthur2:
Hi, @tisonkun , Could I be assigned this issue? Just wanted to gently follow up on this proposal to make sure it wasn't overlooked. Alternatively, should I just go ahead and create a PR directly without an assignment, since this proposal has been discussed before? I'm not sure if you've seen this issue.

tisonkun:
You can directly submit a PR.

But since you're the issue creator and you spend time on this, based on our previous collaboration, I can assign this issue to you.

Typically, I seldom assign issue to certain person unless I know he/she would work on it. Otherwise it's no more than a signal that: "Someone has already taken the issue and I should go away". But often the assignee vanishes later and it prevents others to take it over :D

tisonkun:
@orthur2 This should be resolved by #65?

orthur2:
> [@orthur2](https://github.com/orthur2) This should be resolved by [#65](https://github.com/fast/mea/pull/65)?

Yes, exactly. Since you just merged the fix in #65, we can close this issue now. Thanks.

PR #65, opened 2025-08-10
feat: Implement write lock downgrading

orthur2:
## Description

This PR introduces a `downgrade()` method to `RwLockWriteGuard` and its variants, addressing Issue #62.

Currently, a task holding a write lock must drop it and re-acquire a read lock to allow concurrent reads. This is inefficient and creates a small window for race conditions. The `downgrade()` method provides a way to atomically and efficiently convert an exclusive write lock into a shared read lock, improving performance and ensuring safety.

## Implementation

The `downgrade()` method has been added to all four `RwLockWriteGuard` variants:

- `RwLockWriteGuard`
- `MappedRwLockWriteGuard`
- `OwnedRwLockWriteGuard`
- `OwnedMappedRwLockWriteGuard`

The implementation atomically converts the lock by releasing `max_readers - 1` permits from the underlying semaphore, leaving one permit for the new read guard. `ManuallyDrop` is used to prevent the original write guard's `Drop` implementation from running, ensuring panic safety and correct ownership transfer of the lock.

To implement `RwLockWriteGuard::downgrade`, I'll need to create a new `RwLockReadGuard`. However, the `lock` field of `RwLockReadGuard` is private. To avoid changing the access permissions of the `lock` field directly (given that changing it to `pub(super)` would grant excessive access, and Rust doesn't offer more fine-grained access control, which is a trade-off point tison and I discussed in PR #61), So, my current plan is to add a dedicated constructor for `RwLockReadGuard`. I will name it `from_write_downgrade()` instead of generic `new()` and constrain its API by requiring the `WriteGuard` to be passed in. This makes its purpose clear and prevents misuse.

**This is a Draft Pull Request.**

The implementation of the `downgrade` logic is complete and ready for review. However, this PR cannot be merged yet because a related test, `test_owned_mapped_write_guard_downgrade_no_memory_leak`, is currently failing.

The root cause of this failure is not in the `downgrade` method itself, but in a pre-existing memory leak within the `map` and `filter_map` functions of the `Owned` guard types. This leak is being tracked in Issue #63 and will be resolved by **PR #64**.

Once PR #64 is merged, the tests are expected to pass.


tisonkun:
> Once PR https://github.com/cratesland/mea/pull/64 is merged, the tests are expected to pass.

#64 has been merged. I suppose you can rebase and make this PR ready for review then :D

tisonkun (approved):
LGTM. I made some code style changes and this is ready for merging.

orthur2:
> LGTM. I made some code style changes and this is ready for merging.

Thanks for your help.

PR #66, opened 2025-11-18
feat: impl OnceCell

tisonkun:
cc @orthur2 - You may be interested in taking a look at this PR.

It's a draft PR to implement async `OnceCell` that is an analogy to std's `OnceCell` or `OnceLock` but take a `f: impl FnOnce() -> Fut, Fut: Future<Output = T>` as initiator.

I'm considering whether `OnceCell` or `OnceLock` the more proper name, and you may implement other methods like `get_mut`, `get_mut_or_init`, etc.

### DIFF
--- a/mea/src/lib.rs
+++ b/mea/src/lib.rs
@@ -70,6 +70,7 @@ pub mod condvar;
 pub mod latch;
 pub mod mpsc;
 pub mod mutex;
+pub mod once;
 pub mod oneshot;
 pub mod rwlock;
 pub mod semaphore;
--- a/mea/src/mpsc/bounded.rs
+++ b/mea/src/mpsc/bounded.rs
@@ -27,8 +27,8 @@ use std::task::Poll;
 use std::task::Waker;
 
 use crate::atomicbox::AtomicOptionBox;
+use crate::internal;
 use crate::internal::Acquire;
-use crate::internal::Semaphore;
 use crate::mpsc::error::TrySendError;
 use crate::mpsc::SendError;
 use crate::mpsc::TryRecvError;
@@ -44,7 +44,7 @@ pub fn bounded<T>(buffer: usize) -> (BoundedSender<T>, BoundedReceiver<T>) {
     assert!(buffer > 0, "mpsc bounded channel requires buffer > 0");
     let state = Arc::new(BoundedState {
         senders: AtomicUsize::new(1),
-        tx_permits: Semaphore::new(0),
+        tx_permits: internal::Semaphore::new(0),
         rx_task: AtomicOptionBox::none(),
     });
     let (sender, receiver) = std::sync::mpsc::sync_channel(buffer);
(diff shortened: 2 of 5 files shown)


tisonkun:
> How would you like to make a difference with tokio's `OnceCell`? https://docs.rs/tokio/latest/tokio/sync/struct.OnceCell.html

My initial purpose is to replace `tokio::sync` used in OpenDAL. The internal implementation would be different. tokio's OnceCell depends on a closable semaphore while our semaphore has a different core.

Besides, I can add more methods like `get_mut_or_init` and try to bringing other analogy, e.g., `LazyCell` and `Once`. So you can see I'm naming the module as `once` rather than `once_cell`.

BewareMyPower:
I see. Since the `Semaphore` in mea does not have the `close` method like tokio, the interface is a bit different, e.g. it does not have to handle the case when `acquire` fails.

tisonkun:
> BTW, there are some unnecessary changes that change Semaphore to internal::Semaphore, which violates the idiomatic use paths described in "the book": https://doc.rust-lang.org/book/ch07-04-bringing-paths-into-scope-with-the-use-keyword.html#creating-idiomatic-use-paths

Nope. We have two `Semaphore` in mea, the user-facing one and the low-level internal one. I'd keep the internal one used as `internal::Semaphore` always and thus it's clear which one is used.

orthur2:
Hi @tisonkun, I've taken a closer look and I think there is an issue in the current logic  around semaphore usage .

In `OnceCell::new()`, the semaphore is initialized with 0 permits: `sem: internal::Semaphore::new(0)`, However, inside `get_or_init`, the logic requires acquiring a permit before initialization:  `self.sem.acquire(1).await`.
Since the initial permit count is 0, the first caller will hang here indefinitely waiting for a permit. Because the caller is waiting, they can never execute f().await, and consequently, set_value (which releases the permits) is never reached.

It seems the implementation is treating the semaphore both as a `mutex` and as a `notification` mechanism, which might be problematic. If we want to use the semaphore strictly as a mutex primitive, it probably needs to start with 1 permit.
Alternatively, we could introduce some explicit state values to decouple these responsibilities.

What do you think?

tisonkun:
> In `OnceCell::new()`, the semaphore is initialized with 0 permits: `sem: internal::Semaphore::new(0)`, However, inside `get_or_init`, the logic requires acquiring a permit before initialization: `self.sem.acquire(1).await`. Since the initial permit count is 0, the first caller will hang here indefinitely waiting for a permit. Because the caller is waiting, they can never execute f().await, and consequently, set_value (which releases the permits) is never reached.
> 
> It seems the implementation is treating the semaphore both as a `mutex` and as a `notification` mechanism, which might be problematic. If we want to use the semaphore strictly as a mutex primitive, it probably needs to start with 1 permit. Alternatively, we could introduce some explicit state values to decouple these responsibilities.

Yes correct.

tokio's semaphore can be "closed" and it is used as a mutex and a notifier. In our case, we need some primitives to support:

1. Guard only one actor can try to init the cell.
2. Once the cell inited, notify other participates.

Extra state like an AtomicBool is OK.

PR #67, opened 2025-11-20
chore(semaphore): simplify acquired_or_enqueue

BewareMyPower:
The `acquired` local variable is only updated in https://github.com/fast/mea/blob/451f1b3037e2a2a7b00ad5bd4be1ea2bdca6edb4/mea/src/internal/semaphore.rs#L336

However, after that, there could be only two results:
- if `remaining` is 0: return true directly
- otherwise: break the loop and construct a node whose permits is `needed - acquired` to the wait list

`current.checked_add(acquired)` never makes sense because when the code reaches here, `acquired` is always 0.

This issue should also exit for the tokio's loop here: https://github.com/tokio-rs/tokio/blob/4714ca168d6bd97193625657b0381e9b65a9ceff/tokio/src/sync/batch_semaphore.rs#L416

The key point is that `acquired` is only modified when the CAS operation succeeds but then it will break the loop immediately.

P.S. I realized the issue when I found the `acq` variable is `needed - acquired` in mea but `needed >> Self::PERMIT_SHIFT` in tokio when `total >= needed`. Then I found `acquired` could never be non-zero here.

### DIFF
--- a/mea/src/internal/semaphore.rs
+++ b/mea/src/internal/semaphore.rs
@@ -296,20 +296,20 @@ impl Future for Acquire<'_> {
 
 /// Returns `true` if successfully acquired the semaphore; `false` otherwise.
 fn acquired_or_enqueue(
-    semaphore: &Semaphore,
+    sem: &Semaphore,
     needed: usize,
     idx: &mut Option<usize>,
     waker: Option<&Waker>,
     enqueue_last: bool,
 ) -> bool {
-    let mut current = semaphore.permits.load(Ordering::Acquire);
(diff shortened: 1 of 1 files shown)


tisonkun (approved):
@BewareMyPower Thanks for your contribution! You analyze is correct.

I push a new commit to even further simplify the code:

1. Use guard statement to simplify branches.
2. We don't need to calculate `acquired` because only `remaining` is important.

```
when remaining > 0
acquired = current
remaining = needed - current

in the following used place, we calculate
(needed - acquired) == remaining

this is of course because `permits` of `WaitNode` means the remaining needed permits
```

PR #68, opened 2025-11-20
docs: fix incorrect link for composing delay futures

BewareMyPower:
It seems to be an URL redirection issue?

https://docs.rs/fastimer/*/fastimer/fn.timeout.html will be redirected to https://docs.rs/fastimer/latest/fastimer/fn.timeout.html/, which has a slash suffix that cannot be parsed.

image

Disabling the redirection and removing the slash works.

image


tisonkun:
This ever works and `https://docs.rs/fastimer/*/fastimer` works well. Ditto other links like `https://docs.rs/mea/*/mea/latch/struct.Latch.html`.

> https://docs.rs/fastimer/*/fastimer/fn.timeout.html will be redirected to https://docs.rs/fastimer/latest/fastimer/fn.timeout.html/

Yes. This seems to be a docs.rs bug. You can report it there.

PR #69, opened 2025-11-22
feat: minimum implementation of OnceCell

BewareMyPower:
This PR only includes the very essentially asynchronous implementation of `OnceCell`:
- `new`
- `drop`
- `get`
- `get_or_init`

Some differences from tokio:
- Since mea does not support const version constructors (typically `const_new`), the `OnceCell` in mea only supports shared among various tasks via `Arc<OnceCell<T>>`, while tokio allows the static instance.
- `initialized()` is not exposed, the `get` method should be the alternative to check if the value is initialized with very low overhead (just an construction of `Option` of a reference)

The key implementation is to use a semaphore whose permits is 1, where each `get_or_init` call could acquire 1 permit and release it at the end before `initialized()` returns true. It means there is only 1 task that is able to set the value. Other waiting tasks could be waked by the `release` operation on `Semaphore`.

TODO: once this basic implementation is accepted, I will add more methods based on it.

### DIFF
--- a/mea/src/once/once_cell.rs
+++ b/mea/src/once/once_cell.rs
@@ -55,7 +55,6 @@ unsafe impl<T: Sync + Send> Sync for OnceCell<T> {}
 // SAFETY: OnceCell<T> can be sent between threads as long as T is Send.
 unsafe impl<T: Send> Send for OnceCell<T> {}
 
-
 impl<T> Default for OnceCell<T> {
     fn default() -> Self {
         Self::new()
--- a/mea/src/once/tests.rs
+++ b/mea/src/once/tests.rs
@@ -13,8 +13,9 @@
 // limitations under the License.
 
 use std::sync::Arc;
-use std::sync::atomic::AtomicU32;
+use std::sync::atomic::AtomicBool;
 use std::sync::atomic::Ordering;
+use std::time::Duration;
 
 use tokio::sync::Mutex;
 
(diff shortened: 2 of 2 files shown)


BewareMyPower:
@orthur2 Thanks for your valuable review, I'm addressing the comments (as well as copilot's) soon

BewareMyPower:
@orthur2 Comments should be addressed now, some comment style issues before were mainly related to safety comments. I also added a test for cancellation. PTAL again.

BTW, I found tokio's API comment is wrong. https://docs.rs/tokio/latest/tokio/sync/struct.OnceCell.html#method.get_or_init

If the operation panics, the whole application will panic as well.

orthur2:
> @orthur2 Comments should be addressed now, some comment style issues before were mainly related to safety comments. I also added a test for cancellation. PTAL again.现在应该处理一下评论了，之前的一些评论格式问题主要与安全评论有关。我还添加了取消功能的测试。请再次查看。
> 
> BTW, I found tokio's API comment is wrong. https://docs.rs/tokio/latest/tokio/sync/struct.OnceCell.html#method.get_or_init顺便说一下，我发现 tokio 的 API 注释有误。https [://docs.rs/tokio/latest/tokio/sync/struct.OnceCell.html#method.get_or_init](https://docs.rs/tokio/latest/tokio/sync/struct.OnceCell.html#method.get_or_init)
> 
> If the operation panics, the whole application will panic as well.如果操作失败，整个应用程序也会失败。

Looks great.

PR #70, opened 2025-11-22
condvar missing notify

iGxnon:
This is not an actually PR.

A 1-second synchronous wait was added between releasing the lock on the condvar and entering the wait state. 

During this period, another task might acquire the lock, modify the condition, and signal a wake-up, causing this task to miss the wake-up.

The following test `notify_all` will not reach "HERE 3!".

```
I have start!
I am waiting on it
HERE 1!
HERE 2!
```

This kind of lost wake-up should be unexpected.

Interestingly, this can run normally in the current thread runtime(see `notify_all_local_runtime`) because releasing the lock and entering the wait state are atomic.

### DIFF
--- a/mea/src/condvar/mod.rs
+++ b/mea/src/condvar/mod.rs
@@ -107,6 +107,10 @@ impl Condvar {
     pub async fn wait<'a, T>(&self, guard: MutexGuard<'a, T>) -> MutexGuard<'a, T> {
         let mutex = mutex::guard_lock(&guard);
         drop(guard);
+
+        // sync point
+        std::thread::sleep(std::time::Duration::from_secs(1));
+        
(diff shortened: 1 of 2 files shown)


tisonkun:
This is a known issue for all `notify_all` implementation for all `Condvar`. That a wait can miss a notify_all in some execution order. Using `notify_one` would ensure that a late wait see the permits :D

tisonkun:
Maybe this worth a fix. I think if users are writing code that only call the condvar's notify methods when hold the lock, this fix can ease your original edge cases.

You can try to make a PR and perhaps make `let _ = Pin::new(&mut acquire).poll(&mut Context::from_waker(&Waker::noop()))` more smooth by factoring the logic from `Acquire`.

PR #71, opened 2025-11-23
fix(condvar): make condvar wait() register waiter while holding lock

iGxnon:
See https://github.com/fast/mea/pull/70

### DIFF
--- a/mea/src/condvar/mod.rs
+++ b/mea/src/condvar/mod.rs
@@ -46,6 +46,8 @@
 //! ```
 
 use std::fmt;
+use std::future::poll_fn;
+use std::task::Poll;
 
 use crate::internal;
 use crate::mutex;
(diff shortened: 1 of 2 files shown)


mea/src/condvar/mod.rs
```diff
@@ -106,8 +108,10 @@ impl Condvar {
     /// However, as a best practice avoid using with multiple mutexes.
     pub async fn wait<'a, T>(&self, guard: MutexGuard<'a, T>) -> MutexGuard<'a, T> {
         let mutex = mutex::guard_lock(&guard);
+        // register waiter while holding lock
+        let acquire = poll_fn(|cx| Poll::Ready(self.s.poll_acquire_once(1, cx))).await;
```
k7Xm9pQr, line 112: I cannot use Waker::noop() because it has only been stable since 1.85.0.
Now MSRV is 1.80.

And using noop waiter instead of a wait() context waiter would be a bit strange?

PR #76, opened 2025-11-25
chore: tidy OnceCell code

tisonkun:
This follows up #69


mea/src/once/once_cell.rs
```diff
@@ -46,7 +46,7 @@ use crate::internal;
 pub struct OnceCell<T> {
     value_set: AtomicBool,
     value: UnsafeCell<MaybeUninit<T>>,
-    semaphore: internal::Semaphore,
+    semaphore: Semaphore,
```
tisonkun, line 49: Previously I was considering using the inernal Semaphore's `notify_all` method, but it seems a simply P/V pattern is enough here. Switch to the higher level encapsulation.

Issue #77, opened 2025-11-25
More methods to OnceCell


tisonkun:
* [ ] `get_or_try_init`
* [ ] `get_mut`
* [ ] `get_mut_or_init`
* [ ] `get_mut_or_try_init`
* [ ] `set`
* [ ] `take`

You can refer to:

* https://github.com/tokio-rs/tokio/blob/master/tokio/src/sync/once_cell.rs
* https://github.com/rust-lang/rust/pull/114788

Also, it's possible to add a `LazyCell` primitive that has a pre-configured initializer (closure).

BewareMyPower:
I'm wondering if it's worth implementing a `get_mut_or_init`. This method should accept a mutable reference so that it cannot be called concurrently.

A typical use case for modifying a mutable reference is calling `get_or_init` on an `OnceCell` whose value is something like `Arc<Mutex<T>>` so that it can be modified concurrently.

```rust
#[test]
fn multi_init_mut() {
    let rt = tokio::runtime::Builder::new_multi_thread()
        .worker_threads(4)
        .enable_time()
        .build()
        .unwrap();

    static CELL: OnceCell<AtomicUsize> = OnceCell::new();

    rt.block_on(async {
        const N: usize = 100;

        let mut handles = vec![];
        for i in 0..N {
            handles.push(rt.spawn(async move {
                let result = CELL
                    .get_or_init(move || async move { AtomicUsize::new(1000 + i) })
                    .await;
                result.fetch_add(1, Ordering::Relaxed);
                tokio::time::sleep(Duration::from_millis(100)).await;
                result.load(Ordering::Relaxed)
            }));
        }
        let mut values = vec![];
        for handle in handles {
            values.push(handle.await.unwrap());
        }
        let value = CELL.get().unwrap().load(Ordering::Relaxed);
        assert!(
            value >= 1000 + N && value < 1000 + N + N,
            "final value: {}",
            value
        );
        for v in values {
            assert_eq!(v, value);
        }
    });
}
```

tisonkun:
You can see the linked issue.

While async primitives can be used to share cross threads, our OnceCell can be used as an analogy to std OnceLock while its pass-in function is an async fn.

Thus, this can be used as a cell that initialized only once and need not shared among threads.

tisonkun:
See https://github.com/rust-lang/rust/issues/74465#issuecomment-1676522051

PR #78, opened 2025-11-25
feat: support OnceCell get_or_try_init

tisonkun:
AFAICS OpenDAL would require this method since some of its initializer is faillible.

This refers to #77

### DIFF
--- a/mea/src/once/once_cell.rs
+++ b/mea/src/once/once_cell.rs
@@ -19,6 +19,7 @@ use std::sync::atomic::AtomicBool;
 use std::sync::atomic::Ordering;
 
 use crate::semaphore::Semaphore;
+use crate::semaphore::SemaphorePermit;
 
 /// A thread-safe cell which can nominally be written to only once.
 ///
@@ -106,7 +107,7 @@ impl<T> OnceCell<T> {
             return v;
         }
 
-        let _permit = self.semaphore.acquire(1).await;
+        let permit = self.semaphore.acquire(1).await;
 
         if let Some(v) = self.get() {
             // double-checked: another task initialized the value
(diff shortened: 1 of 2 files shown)


mea/src/once/once_cell.rs
```diff
@@ -126,16 +172,11 @@ impl<T> OnceCell<T> {
...
-    // SAFETY: Caller must ensure that `initialized()` returns true before calling this method.
-    unsafe fn get_unchecked(&self) -> &T {
-        unsafe { &*(*self.value.get()).as_ptr() }
-    }
 }
 
 impl<T> Drop for OnceCell<T> {
     fn drop(&mut self) {
-        if self.initialized() {
+        if *self.value_set.get_mut() {
```
BewareMyPower, line 179: Why did you make this change?

tisonkun: Rust would guarantee that no other threads are concurrently accessing the atomic data if we have a `&mut` ref.

Actually, the memory barrier will be inserted by the compiler and there maybe further improvement under this encapsulation.

BewareMyPower: > the memory barrier will be inserted by the compiler

Is there any official Rust document for this? It's okay that no other threads are concurrently accessing the atomic data. My concern is that if the value (`self.value`) updated before atomic release store (on `self.value_set`) visible in the destructor without atomic acquire read (on `self.value_set`)?

BewareMyPower: image

I talked to Gemini just now, but not sure if Gemini is misleading me. :( so it would be helpful if we can clarify it with some official documents

BewareMyPower: Ah okay, I corrected Gemini now.

image

image

PR #80, opened 2025-11-25
feat: impl try_insert and set for OnceCell

BewareMyPower:
main issue: #77

### DIFF
--- a/mea/src/once/once_cell.rs
+++ b/mea/src/once/once_cell.rs
@@ -207,10 +207,6 @@ impl<T> OnceCell<T> {
         }
 
         let optional_permit = self.semaphore.try_acquire(1);
-        println!(
-            "try_insert acquired permit: {:?}",
-            optional_permit.is_some()
-        );
         match self.get() {
             // double-checked: another task initialized the value
             Some(v) => Err((Some(v), value)),


mea/src/once/once_cell.rs
```diff
@@ -182,6 +183,34 @@ impl<T> OnceCell<T> {
...
+    /// Sets the value of the cell to the given value if `OnceCell` is empty.
+    ///
+    /// If the cell is already initialized, returns a [`SetError::AlreadyInitializedError`].
+    ///
+    /// If the cell is currently being initialized by another task, returns a
+    /// [`SetError::InitializingError`].
+    ///
+    /// [`SetError::AlreadyInitializedError`]: crate::once::SetError::AlreadyInitializedError
+    /// [`SetError::InitializingError`]: crate::once::SetError::InitializingError
+    pub fn set(&self, value: T) -> Result<(), SetError<T>> {
```
tisonkun: I'd prefer a signature like https://doc.rust-lang.org/std/sync/struct.OnceLock.html#method.set

That is, no need for an extra `SetError` type. Only return the passed-in value when it's not initialized by the passed-in value.

mea/src/once/tests.rs
```diff
@@ -158,3 +159,35 @@ async fn init_error() {
...
+            })
+            .await;
+        assert_eq!(*value, 3);
+    });
+
+    let handle2 = tokio::spawn(async {
+        tokio::time::sleep(Duration::from_millis(50)).await;
+        let error = CELL.set(4).unwrap_err();
+        println!("Error: {} (debug: {:?})", &error, &error);
+        assert!(error == SetError::InitializingError(4));
```
tisonkun: You can surely use `assert_eq` for the same purpose. It would print the value when assertion failed.

tisonkun:
Besides, `set` can be a special case of `try_insert` if you take a look at https://doc.rust-lang.org/std/sync/struct.OnceLock.html#method.try_insert and it's impls.

mea/src/once/once_cell.rs
BewareMyPower: It makes sense. I can also hardly to think of a case that callers need to handle different types of `SetError`.

BewareMyPower:
Yes, `set` is much harder to use than `get_or_init` in asynchronous cases.

Before checking tokio's implementation, I thought `cell.set(value)` is similar to `cell.get_or_init(async move { value })`.

```rust
    pub async fn set(&self, value: T) -> Result<&T, T> {
        if self.is_initialized() {
            return Err(value);
        }
        let permit = self.semaphore.acquire(1).await;
        if self.is_initialized() {
            return Err(value);
        }
        Ok(self.set_value(value, permit))
    }
```

The difference is that, when the value to set is not lazily initialized,
- with `get_or_init`, the ownership of `value` is moved to the `init` function and there is no way to get it again
- with `set`, the ownership of `value` will be moved back to `Err(value)`, so users can process the failed values

The question is do we really need to support this case?

PR #82, opened 2025-11-29
chore: fine tune OnceCell impls

tisonkun:


mea/src/once/once_cell.rs
```diff
@@ -270,24 +263,23 @@ impl<T> OnceCell<T> {
         F: FnOnce() -> Fut,
         Fut: Future<Output = Result<T, E>>,
     {
-        if self.is_initialized() {
+        // Workaround if let Some(v) = self.get_mut() { return Ok(v); }
+        // @see https://github.com/rust-lang/rust/issues/51545
+        if self.initialized_mut() {
+            // SAFETY: checked is_initialized and we have a unique access
```
BewareMyPower: `is_initialized` does not match the actual method name

mea/src/once/once_cell.rs
```diff
@@ -220,16 +214,15 @@ impl<T> OnceCell<T> {
...
-            let value = init().await;
-            // SAFETY: we have exclusive access via holding &mut self
-            unsafe {
-                self.set_value(value);
-                self.get_unchecked_mut()
-            }
+        // Workaround if let Some(v) = self.get_mut() { return v; }
+        // @see https://github.com/rust-lang/rust/issues/51545
+        if self.initialized_mut() {
+            // SAFETY: checked is_initialized and we have a unique access
```
BewareMyPower: The same with previous comment. I think we'd better not mention the method name directly in comments. Instead, use "initialization" or some other natural language for it.

mea/src/once/once_cell.rs
tisonkun: Good catch.

mea/src/once/once_cell.rs
tisonkun: Reasonable.

PR #83, opened 2025-11-29
feat: add mutations to OnceCell

tisonkun:
* This refers to #77.
* This closes #80.

### DIFF
--- a/mea/src/once/once_cell.rs
+++ b/mea/src/once/once_cell.rs
@@ -273,7 +273,7 @@ impl<T> OnceCell<T> {
     /// Initializes the contents of the cell to `value` if the cell was uninitialized,
     /// then returns a reference to it.
     ///
-    /// May pending if another thread is currently attempting to initialize the cell. The cell is
+    /// May wait if another thread is currently attempting to initialize the cell. The cell is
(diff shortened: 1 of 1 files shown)


mea/src/once/once_cell.rs
```diff
@@ -270,6 +270,168 @@ impl<T> OnceCell<T> {
...
+    ///     assert_eq!(CELL.set(92).await, Ok(()));
+    /// });
+    ///
+    /// assert_eq!(CELL.wait().await, &92);
+    /// # }
+    /// ```
+    pub async fn wait(&self) -> &T {
+        while !self.initialized() {
+            let _ = self.semaphore.acquire(1).await;
+        }
```
tisonkun: Since our use case of semaphore is an exclusive lock, I'm considering replace the semaphore with `Mutex` + `Condvar` pair.

tisonkun: But there is a risk of missing a notification.

tisonkun: Maybe use a count down latch (with value = 1).

tisonkun: Or we can remove this method now.

PR #86, opened 2025-11-30
feat(once): implement Once primitive

orthur2:
This PR implements a lock-free fast path and an async-mutex–based slow path for a Once primitive. The design is mainly inspired by Rust’s std::sync::Once and Go’s sync.Once  which I’m familiar with.

I didn’t create a dedicated once.rs file because it would conflict with the module name, so I placed the implementation directly inside mod.rs.
At the moment, wait() (similar to Rust’s Once::wait()) is not implemented yet. If we decide to add it, I think it would be better to use CountdownState instead of AtomicBool, so we can properly support wait() semantics.
The README.md has not been updated to reflect the new once module yet.

Please help me review the code and point out any mistakes. If you have any additional thoughts or suggestions, I’d greatly appreciate them.

### DIFF
--- a/mea/src/once/mod.rs
+++ b/mea/src/once/mod.rs
@@ -12,10 +12,185 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-//! Asynchronous primitives for one-time initialization.
+//! Asynchronous primitives for one-time async coordination.
+//!
+//! The module currently provides:
+//!
+//! - [`Once`]: An async synchronization primitive that ensures a one-time asynchronous operation
+//!   runs at most once, even when called concurrently.
+//! - [`OnceCell`]: A cell that can be written to only once, storing a value produced
+//!   asynchronously.
+
+use std::fmt;
+use std::ops::AsyncFnOnce;
+use std::sync::atomic::AtomicBool;
+use std::sync::atomic::Ordering;
+
+use crate::semaphore::Semaphore;
 
 mod once_cell;
+
 pub use self::once_cell::OnceCell;
 
 #[cfg(test)]
 mod tests;
+
+/// A synchronization primitive which can be used to run a one-time async initialization.
+///
+/// Unlike [`std::sync::Once`], this type never blocks a thread. The provided closure must
+/// produce a future and the future is awaited inside the primitive. Coordination happens
+/// with asynchronous semaphores, which keeps the implementation runtime-agnostic.
+///
+/// This type also intentionally omits "poisoning" semantics. If an initialization future is
+/// cancelled or panics, the attempt is abandoned and other tasks may retry the operation.
+/// Encode partial-initialization detection in the future itself (e.g. return a `Result`)
+/// when needed.
+///
+/// This type can only be constructed with Once::new().
+///
+/// See the [module level documentation](crate::once) for additional context.
+///
+/// # Examples
+///
+/// ```
+/// # #[tokio::main]
+/// # async fn main() {
+/// use std::sync::atomic::AtomicUsize;
+/// use std::sync::atomic::Ordering;
+///
+/// use mea::once::Once;
+///
+/// static ONCE: Once = Once::new();
+/// static COUNTER: AtomicUsize = AtomicUsize::new(0);
+///
+/// let handle1 = tokio::spawn(async {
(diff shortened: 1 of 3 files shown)


mea/src/once/mod.rs
```diff
@@ -12,10 +12,186 @@
...
+    /// static ONCE: Once = Once::new();
+    ///
+    /// assert!(!ONCE.is_completed());
+    ///
+    /// ONCE.call_once(|| async {}).await;
+    ///
+    /// assert!(ONCE.is_completed());
+    /// # }
+    /// ```
+    #[inline]
```
BewareMyPower: I'm wondering when should `#[inline]` be used rather than `#[inline(always)]`.

mea/src/once/tests.rs
```diff
@@ -13,55 +13,16 @@
...
-}
-
-impl Drop for Foo {
-    fn drop(&mut self) {
-        self.value.store(true, Ordering::Release);
-    }
-}
-
-#[tokio::test]
-async fn drop_cell() {
```
BewareMyPower, line 48: Why did you remove this test? This test can guarantee the underlying `T` will be dropped if the `OnceCell<T>` is dropped.

mea/src/once/tests.rs
```diff
@@ -200,3 +161,136 @@ async fn get_mut_or_try_init() {
...
 }
+
+#[tokio::test]
+async fn test_call_once_runs_only_once() {
+    static ONCE: Once = Once::new();
+    static COUNTER: AtomicUsize = AtomicUsize::new(0);
+
+    assert!(!ONCE.is_completed());
+
+    ONCE.call_once(|| async {
```
BewareMyPower: We can use async closures instead. See https://github.com/fast/mea/pull/84

mea/src/once/tests.rs
orthur2: You are right. Removing the drop_cell test was not intentional.

Originally I added the new tests at the top of test.rs because I was thinking in the order of “Once tests before OnceCell tests” (since this is the once module). Later I decided it would be cleaner to append new tests at the end of the file instead. During the move/refactor I accidentally deleted the test, and when reviewing my changes before committing I only checked that the test suite passed, but unfortunately overlooked the modifications in the test file itself.

Sorry for the confusion. This was entirely my mistake caused by carelessness during the test rearrangement. I will restore the test.

mea/src/once/mod.rs
orthur2: Since I rarely write C++-style ultra-low-level performance-critical code, I'll just share my current understanding which may by wrong.

I don't think #[inline(always)] is necessary here. In most cases the compiler is able to make better inlining decisions than we can, and #[inline] already provides enough flexibility for the optimizer. My impression is that #[inline(always)] is more appropriate only in very extreme hot-path scenarios where benchmarks clearly show that forcing inlining provides measurable benefits.If we later find that this fast path is hot enough to justify always, I'm open to reconsidering it. But for now I prefer the conservative default.

BewareMyPower: I mean, if you're really going to tell the compiler to make the function inline, just use `#[inline(always)]`. Otherwise, never write `#[inline]`. It's just a hint that you never know whether it works even that the compiler might just ignore this keyword when perform optimization. For example, GCC just ignores the `inline` keyword. See https://www.kernel.org/doc/local/inline.html

> Current versions of gcc turned "inline" into a request (similar to the old "register" keyword), rendering it essentially useless.

IMO, `#[inline]` is a bad design in Rust. It just confuse users that simple methods (especially an one-line getter implementation) could be optimized with this keyword. The reality is that compiler is far more clever to make an one-line implementation inline.

P.S. In C++, `inline` is only used for [ODR](https://en.wikipedia.org/wiki/One_Definition_Rule) now.

orthur2: Thanks for the detailed explanation and the kernel doc reference — that makes sense. I learned something from this. I’ll update the PR accordingly.

tisonkun (approved):
Interesting.

I'm going to merge this first. But I'm considering to implement `Once` with `CountDownState` instead of `AtomicBool` to support `wait`. And then `Once` can be the internal of `OnceCell`.

We can do that in some follw-ups, though.

PR #90, opened 2025-12-07
feat: impl broadcast channel

### LINKED ISSUE #88: Support broadcast channel
# Motivation

I'm going to factor out the async runtime code in [foyer](https://github.com/foyer-rs/foyer) (cc @MrCroxx) to make `madsim` an opt-outable feature.

Typically, it would be like a [runtime](https://github.com/scopedb/percas/blob/89e23e9e46d2ee169f44e0542f2e2ddae8fafe62/crates/core/src/runtime.rs) module to hide the implementation.

Before that, I noticed foyer uses several `tokio::sync` primitives which can be replaced with `mea`:

* `tokio::sync::Barrier` can be `mea::barrier::Barrier`
* `tokio::sync::oneshot` can be `mea::oneshot`
* `tokio::sync::Mutex` can be `mea::mutex::Mutex`
* `tokio::sync::broadcast` is lacking

Thus, I think we can support the broadcast channel in mea to replace all `tokio::sync` usage in foyer.

# Design and Implementation

The most common design is to use a circular queue, where senders put elements at the end, and each receiver tracks its current position and consumes new elements if any are available.

Some decision points:

1. What if a receiver lags too much until the next new elements put would overflow its current position?

tokio's broadcast would trigger a [`Lagged`](https://docs.rs/tokio/latest/tokio/sync/broadcast/#lagging) error and forward the receiver's position to the oldest value contained by the channel.

[async-broadcast](https://docs.rs/async-broadcast/latest/async_broadcast/) supports different [overflow policy](https://docs.rs/async-broadcast/latest/async_broadcast/struct.Sender.html#method.overflow): (1) like what tokio's broadcast does, (2) sender waits for all the receivers to move forward.

2. Then we need to decide the API structure, like:

* How to make a broadcast channel?
* How to make a new sender/receiver of the broadcast channel?
* How sender's/receiver's (try) send/recv signatures look like, esp. sync/async. faillible, etc.

3. How to bridge Rust's `Future` & `Waker` mechanism?

Typically, each receiver would consume elements until there are no more and register a waker on that cell. Once a sender puts a new element in that cell, it can wake up all the wakers so that receivers can consume the new elements.

The main point here is how we can efficiently mutate the shared circular queue and reduce lock usage as much as possible.

Besides, if we allow sender to wait on channel full, we may need to register wakers for wating senders.

# Reference

* https://docs.rs/async-broadcast/latest/async_broadcast/index.html
* https://docs.rs/tokio/latest/tokio/sync/broadcast/index.html
* https://docs.rs/postage/0.5.0/postage/broadcast/index.html
* https://docs.rs/broadcaster/1.0.0/broadcaster/

cc @BewareMyPower @orthur2

tisonkun:
This closes #88 

I'd first implement an overflow policy to overwrite the oldest elements and let receivers be lagged.

We may later add more strategies and shift down the current impl to `mea::broadcast::overflow::channel`. But let's avoid to do that unless we find it really needed.

Besides, we can change the `channel` argument `capacity` to `NonZeroUsize` when we are making breaking changes.

### DIFF
--- a/mea/src/internal/rwlock.rs
+++ b/mea/src/internal/rwlock.rs
@@ -31,3 +31,23 @@ impl<T: ?Sized> RwLock<T> {
         self.0.write().unwrap_or_else(PoisonError::into_inner)
     }
 }
+
+#[cfg(test)]
+mod tests {
+    use std::sync::Arc;
+
+    use crate::internal::RwLock;
+
+    #[test]
+    fn test_poison_rwlock() {
+        let rwlock = Arc::new(RwLock::new(42));
+        let r = rwlock.clone();
+        let handle = std::thread::spawn(move || {
+            let _guard = r.write();
+            panic!("poison");
+        });
+        let _ = handle.join();
+        assert_eq!(*rwlock.read(), 42);
+        assert_eq!(*rwlock.write(), 42);
+    }
+}


mea/src/broadcast/tests.rs
```diff
@@ -0,0 +1,214 @@
...
+    assert_eq!(rx.recv().await, Ok(1));
+
+    tx.send(2);
+    tx.send(3);
+    tx.send(4);
+    tx.send(5);
+
+    assert_eq!(rx.recv().await, Ok(2));
+    // FIXME: wrapping just hit the head, but this should be lagged
+    assert_eq!(rx2.recv().await, Ok(4));
```
tisonkun: We need to figure out what to do with this case. May or may not be able to work it out.

Perhaps a better "version" strategy and a proper forward solution.

mea/src/broadcast/mod.rs
```diff
@@ -0,0 +1,501 @@
...
+struct Recv<'a, T> {
+    receiver: &'a mut Receiver<T>,
+    idx: Option<usize>,
+}
+
+impl<T: Clone> Future for Recv<'_, T> {
+    type Output = Result<T, RecvError>;
+
+    // TODO: try to leverage try_recv so that we're sure the implement is the same
+    fn poll(self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Self::Output> {
```
tisonkun: Before we merge this patch, we'd better avoid double impl.

mea/src/broadcast/tests.rs
tisonkun: See the new note. Always use `u64` as the version field so it won't be overflow in a reasonable real-world use case.

mea/src/broadcast/tests.rs
```diff
@@ -0,0 +1,211 @@
...
+    assert_eq!(rx2.recv().await, Ok(3));
+}
+
+#[tokio::test]
+async fn test_overflow() {
+    let (tx, mut rx) = channel(4);
+    let mut rx2 = rx.clone();
+
+    let boundary = u64::MAX - 2;
+    tx.shared.tail_cnt.store(boundary, Ordering::SeqCst);
```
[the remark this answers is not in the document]
tisonkun: This is intented to be a white-box test.

mea/src/broadcast/mod.rs
```diff
@@ -0,0 +1,477 @@
...
+                    return Ok(msg.clone());
+                }
+            }
+
+            // If version != head, the slot was overwritten.
+            // This means we lagged, but the `diff > cap` check missed it (likely due to overflow
+            // wrapping). We treat this as a lag.
+            drop(slot);
+
+            let missed = tail.wrapping_sub(self.head).wrapping_sub(cap);
```
orthur2, line 391: If slot.version != head (which means a new message was sent), we must reload the latest tail. Otherwise, missed can underflow and head will also be incorrect. I think we need to re-read tail, recompute the diff, and re-check the condition.

mea/src/broadcast/mod.rs
```diff
@@ -0,0 +1,477 @@
...
+    /// # Examples
+    ///
+    /// ```
+    /// use mea::broadcast;
+    ///
+    /// let (tx, mut rx) = broadcast::channel(16);
+    /// tx.send(10);
+    /// assert_eq!(rx.try_recv(), Ok(10));
+    /// ```
+    pub fn send(&self, msg: T) {
```
orthur2, line 246: In send function, we `fetch_add(1)` to bump `tail_cnt` and effectively publish a new position before writing the `slot`’s data/version. During that window a receiver can observe the new `tail_cnt (diff > 0)`, read the slot, and find it hasn’t been written yet (old version or msg still None), which is currently treated as overwritten and can trigger errors. Would writing the slot first and only then publishing the new position be better?

orthur2: After thinking more carefully, I believe my previous idea might have been wrong. Writing the slot first and then updating `tail_cnt` is not a good improvement. Because the implementation becomes more complex, and it also introduce blocking. 
A more elegant solution I can thinkof is to keep the sender logic unchanged, and instead handle the “data not ready yet” case on the receiver side. By comparing slot.version with head, we can distinguish three states: `slot.version < head` means the slot hasn’t been written yet, `slot.version == head` is the readable case, and `slot.version > head` is the real lagged case. To avoid overflow issues, the comparison should use a wrapping diff (e.g., `version_diff = head.wrapping_sub(slot.version)` ) rather than direct < and >.”

PR #93, opened 2025-12-09
feat: impl LazyCell

tisonkun:
But since the init fn is async, we can't implement `Deref` like `std::sync::LazyCell`.

I wonder whether we really need LazyCell, but at least it allows users to define the initiator at the ctor, rather than always passing in later in `OnceCell::get_or_init(&self, f)`.

### DIFF
--- a/mea/src/once/lazy_cell.rs
+++ b/mea/src/once/lazy_cell.rs
@@ -1,0 +1,93 @@
+// Copyright 2024 tison <wander4096@gmail.com>
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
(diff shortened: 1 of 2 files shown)


mea/src/once/lazy_cell.rs
```diff
@@ -0,0 +1,93 @@
...
+    pub fn get(&self) -> Option<&T> {
+        if self.initialized() {
+            Some(unsafe { &(*self.data.get()).value })
+        } else {
+            None
+        }
+    }
+
+    ///
+    pub async fn get_or_init(&self) -> &T {
```
orthur2, line 61: In get_or_init, ManuallyDrop::take moves the initializer closure before the .await. If the future is cancelled or panics, the closure has already been taken, but value_set is still false. I think there is a logical flaw.

PR #94, opened 2025-12-12
fix(broadcast): fix and test try_recv

orthur2:
Following the comments in PR #90, change the logic in `broadcast::Receiver::try_recv` to be based on the wrapping version difference (`version_diff = head.wrapping_sub(slot.version)`). If the version matches and a message is present, return it normally. If the slot version is logically newer than head, re-read the latest tail and only report Lagged after confirming that the slot has indeed been overwritten.The slot version is initialized to 0, and combined with `msg == None` to indicate the `not ready yet` state.

Add tests to verify that when tail is advanced but the slot has not finished writing, `try_recv` returns Empty instead of incorrectly reporting Lagged.

By the way, while fixing the broadcast implementation, I realized that if a task is canceled while waiting (for example, due to `timeout(recv)`), its Waker may remain in the WaitSet. If the primitive does not get triggered for a long time (e.g. a long-idle broadcast channel, or a barrier that never reaches the final participant), these `zombie wakers` may never be cleaned up. In practice this is usually not a problem because events tend to be triggered frequently, but do we need to consider adjusting the `WaitSet` API for such extreme cases (for example, providing an `unregister method and calling it in `Drop`)?

### DIFF
--- a/mea/src/broadcast/mod.rs
+++ b/mea/src/broadcast/mod.rs
@@ -377,17 +377,19 @@ impl<T: Clone> Receiver<T> {
             let slot = shared.buffer[idx].read();
 
             if slot.version == head {
-                if let Some(msg) = &slot.msg {
+                return if let Some(msg) = &slot.msg {
                     self.head = head.wrapping_add(1);
-                    return Ok(msg.clone());
-                }
+                    Ok(msg.clone())
+                } else {
+                    Err(TryRecvError::Empty)
+                };
             }
+
+            drop(slot);
 
             // If version != head, the slot was overwritten.
             // This means we lagged, but the `diff > cap` check missed it (likely due to overflow
             // wrapping). We treat this as a lag.
-            drop(slot);
-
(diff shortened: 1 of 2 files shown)


tisonkun:
> but do we need to consider adjusting the WaitSet API for such extreme cases

Yeah. This is possible and helps in improving `Barrier`/`WaitGroup`/`Latch`. Previously, I didn't write it because I didn't (and all these waiters are supposed to be finished in a reasonable time. But anyway you can follow what `Acquire` in `internal::semaphore` does to do a deregister.

orthur2:
> Can we fix this by adding:
> 
> ```rust
>             if slot.version == head {
>                 return if let Some(msg) = &slot.msg {
>                     self.head = head.wrapping_add(1);
>                     Ok(msg.clone())
>                 } else {
>                     Err(TryRecvError::Empty)
>                 }
>             }
> ```
> 
> ?

I' ve updated the fix accordingly and re-committed the changes based on your proposal. I also fixed a small typo (resv) and a minor whitespace issue in the documentation in the process.

Issue #91, opened 2025-12-08
Unified TryRecvError & RecvError among channels


tisonkun:
Related features - https://github.com/fast/mea/issues/88#issuecomment-3624054769

Now we have `oneshot` / `mpsc` / `broadcast` channels. And we handle recv error quite differently:

* Whether a RecvError / TryRecvError
* Whether to use `Option` to denote the empty state.

We'd better unify the usage and thus keep the interface consistent.

NOTE: This would be a breaking change.

orthur2:
I've surveyed the current state of Tokio's channel APIs across different channel types:

*   **oneshot**
    *   `Receiver::recv().await`: `Result<T, RecvError>` (indicates the sender was dropped / canceled)
    *   `try_recv`: `Result<T, TryRecvError>` with variants `Empty` | `Closed`
*   **mpsc**
    *   `Receiver::recv`: `Option<T>`, where `None` indicates the channel is closed
    *   `try_recv`: `Result<T, TryRecvError>` with variants `Empty` | `Disconnected`
    *   Notably, there is no `RecvError` type for mpsc
*   **broadcast**
    *   `Receiver::recv`: `Result<T, RecvError>` with variants `Closed` | `Lagged(u64)`
    *   `try_recv`: `Result<T, TryRecvError>` with variants `Empty` | `Closed` | `Lagged(u64)`

So even in Tokio itself, the APIs are not fully unified. For example, `mpsc::recv` uses `Option`, while `oneshot` and `broadcast` use `Result`. Since Tokio itself is inconsistent (mixing `Option` and `Result` across channels), simply mirroring it won't solve the problem.

First and foremost, blocking `recv` only concerns itself with "successfully retrieving a value" or "channel termination". Therefore, its error type should only contain unrecoverable termination states.
Non-blocking `try_recv` needs to express the "temporarily empty" state. Thus, it requires a separate `TryRecvError` containing `Empty` (temporarily no message), `Closed` (permanently no message), and for broadcast, `Lagged(u64)`. Mixing these states into a single error type would force blocking callers to handle impossible branches like "temporarily empty".

Additionally, `broadcast` channels have a special error state: `Lagged`. Furthermore, `Option::None` currently carries ambiguous meanings: in `mpsc::recv` it means "Permanently Closed", while in `oneshot::try_recv` it means "Temporarily Empty". This confuses users. Therefore, I believe dropping `Option` is the better choice.

[...]

PR #98, opened 2025-12-22
refactor(channels): unify channel  errors across oneshot/mpsc/broadcast

orthur2:
This PR originates from issue #91 and standardizes the channel interfaces: All recv() methods now return Result<_, RecvError>, and all try_recv() methods return Result<_, TryRecvError>. The closed state is uniformly represented as Disconnected.
- mpsc: recv no longer returns Option; it now returns Result<T, RecvError::Disconnected>. The error types are exposed as mpsc::RecvError and mpsc::TryRecvError.
- oneshot: try_recv is changed to return Result<T, TryRecvError>, with empty and disconnected states distinguished.
- broadcast: Documentation and error naming are aligned with the other channel types.

**Breaking change**: The API shape has changed; callers must handle the new error types and return values.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -47,7 +47,7 @@ jobs:
       - uses: taiki-e/install-action@v2
         with:
           tool: typos-cli,taplo-cli,hawkeye
-      - run: cargo +nightly x lint
+      - run: cargo x lint
 
   test:
     name: Run tests
--- a/xtask/src/main.rs
+++ b/xtask/src/main.rs
@@ -63,7 +63,7 @@ struct CommandTest {
 
 impl CommandTest {
     fn run(self) {
-        run_command(make_test_cmd(self.no_capture, true, &[]));
+        run_command(make_test_cmd(self.no_capture, &[]));
     }
 }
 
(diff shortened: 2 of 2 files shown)


mea/src/broadcast/mod.rs
```diff
@@ -125,7 +125,7 @@ pub fn channel<T: Clone>(capacity: usize) -> (Sender<T>, Receiver<T>) {
 }
 
 /// Error returned by [`Receiver::recv`].
-#[derive(Debug, Clone, PartialEq, Eq)]
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
```
tisonkun: Rust may typically not assume an `Error` is `Copy` even if it can be.

mea/src/mpsc/error.rs
```diff
@@ -109,6 +109,21 @@ impl<T> fmt::Debug for TrySendError<T> {
 
 impl<T> std::error::Error for TrySendError<T> {}
 
+/// Error returned by `recv`.
+#[derive(PartialEq, Eq, Clone, Copy, Debug)]
```
tisonkun: Ditto. Or why would you like to have these `RecvError`s `Copy`?

orthur2: I added Copy mainly for convenience when handling errors (e.g., matching）and sorry that I’m not very familiar with Rust error best practices yet. Thanks for the reminder. I’ll keep learning and try to follow the conventions better in the future.

tisonkun: I actually found std mpsc derive Copy for `RecvError`, but I'd postpone such a feature until some users find it useful. Copy is a strict constraint that all fields are trivial and that the struct can be copied in a bit-for-bit fashion. I'm not sure we'd keep these Errors in these constraints, at least for now.

mea/src/oneshot/mod.rs
tisonkun, line 1: oneshot's `SendError` doesn't implement PartialEq/Eq/Clone because handling NonNull can be tricky. Typically users can use `into_inner` and use the value `T`.

If it's really desired, we can add the impl which is not a breaking change.

PR #100, opened 2025-12-30
refactor(oneshot): implement `Debug` manually

Paraworker:
Replace derived `Debug` with a manual implementation to avoid an unnecessary `T: Debug` bound.


tisonkun:
Let me know if you're waiting for a new release. The next release would be 0.6.0 and I'd do a review for any other pending break changes.

Also I'm glad to know if you're using `mea` in any of your projects @Paraworker.

PR #111, opened 2026-05-13
fix: tighten oneshot memory ordering

tisonkun:
## Summary
- tighten oneshot sender/receiver state transitions to synchronize final channel ownership handoff
- avoid racing with the sender when dropping a pending async receiver by first moving RECEIVING back to EMPTY
- add concurrent send/drop with try_recv/poll regression tests inspired by upstream oneshot Miri coverage

## Verification
- cargo x test
- cargo +nightly x lint

### DIFF
--- a/mea/src/oneshot/mod.rs
+++ b/mea/src/oneshot/mod.rs
@@ -114,10 +114,6 @@ unsafe impl<T: Sync> Sync for Sender<T> {}
 
 #[inline(always)]
 fn sender_wake_up_receiver<T>(channel: &Channel<T>, state: u8) {
-    // ORDERING: Synchronizes with writing waker to memory, and prevents the
-    // taking of the waker from being ordered before this operation.
-    fence(Ordering::Acquire);
-
     // Take the waker, but critically do not awake it. If we awake it now, the
     // receiving thread could still observe the AWAKING state and re-await, meaning
     // that after we change to the MESSAGE state, it would remain waiting indefinitely
@@ -125,7 +121,8 @@ fn sender_wake_up_receiver<T>(channel: &Channel<T>, state: u8) {
     //
     // SAFETY: at this point we are in the AWAKING state, and the receiving thread
     // does not access the waker while in this state, nor does it free the channel
-    // allocation in this state.
+    // allocation in this state. The caller's acquire ordering establishes a happens-before
+    // relationship with the writing of the waker.
     let waker = unsafe { channel.take_waker() };
 
     // ORDERING: this ordering serves two-fold: it synchronizes with the acquire load
@@ -171,11 +168,11 @@ impl<T> Sender<T> {
         // * RECEIVING + 1 = AWAKING
         // * DISCONNECTED + 1 = EMPTY (invalid), however this state is never observed
         //
-        // ORDERING: we use release ordering to ensure writing the message is visible to the
-        // receiving thread. The EMPTY and DISCONNECTED branches do not observe any shared state,
-        // and thus we do not need an acquire ordering. The RECEIVING branch manages synchronization
-        // independent of this operation.
-        match channel.state.fetch_add(1, Ordering::Release) {
+        // ORDERING: we need release ordering to allow the receiver to synchronize with our write
+        // of the message and with our final write of the state, in the case where the receiver
+        // becomes responsible for freeing the channel. We need acquire ordering in the RECEIVING
+        // and DISCONNECTED branches, as explained further down.
+        match channel.state.fetch_add(1, Ordering::AcqRel) {
             // The receiver is alive and has not started waiting. Send done.
             EMPTY => Ok(()),
             // The receiver is waiting. Wake it up so it can return the message.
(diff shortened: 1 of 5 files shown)


orthur2:
Memory ordering here is tough to think through and guarantee correctness, so my review may not be very comprehensive.

First and foremost, following Copilot's review, `Recv::drop` needs acquire synchronization before dropping the stored waker. `Channel::write_waker` publishes the waker with a Release CAS, but the new `RECEIVING -> EMPTY` CAS in `Recv::drop` is Relaxed. Since `Recv<T>` is Send, the future may be polled on one thread and dropped on another, so Relaxed is too weak here. And the `Recv::poll` waker-replacement path also has the same issue.

Then, I found an issue that existed before this PR. I am concerned that  the `AWAKING` branch in `Recv::drop` can still deallocate the channel too early. This path writes `DISCONNECTED` before it handles `AWAKING`. So if the sender has just moved the state to `AWAKING` and is still taking the waker, the drop path can read back its own `DISCONNECTED` write and free the channel too early.

As Copilot noted, the new spin-loop tests should also be bounded with a timeout or iteration budget.

mea/src/oneshot/mod.rs
```diff
@@ -503,22 +536,43 @@ impl<T> Drop for Recv<T> {
...
+            // SAFETY: The RECEIVING state guarantees we have written a waker.
+            unsafe { channel.drop_waker() };
+        }
+
+        // Set the channel state to disconnected and read what state the channel was in.
+        //
+        // ORDERING: Release is required so that in the states where the sender becomes responsible
+        // for deallocating the channel, they can synchronize with this final state write from us.
+        // Acquire is required by the branches below to synchronize with writes from the sender.
+        match channel.state.swap(DISCONNECTED, Ordering::AcqRel) {
```
orthur2: I think this can free the channel too early. When this `swap()` sees `AWAKING`, it also writes `DISCONNECTED`, so the loop below may just read our own write back and stop waiting while the sender is still taking the waker or publishing the final state.

orthur2:
I pushed a follow-up commit to address it.

orthur2 (approved):
LGTM from my side.

tisonkun:
I'm reviewing the comments and order. From agent review, it says:

Blockers:
1. Please explain the Recv::drop interleaving around the RECEIVING -> EMPTY -> DISCONNECTED path. In particular: if a sender concurrently succeeds in send during the temporary EMPTY window, who owns dropping the message, the waker, and the allocation, and in what order? The current code may rely on the next loop iteration doing EMPTY -> DISCONNECTED, but that safety argument is not written down clearly enough for this kind of code.
2. Fix the misleading comments in oneshot/mod.rs, especially the poll MESSAGE branch saying the sender was dropped when it actually sent a message, and the stale memory-barrier comment around RECEIVING -> EMPTY. In unsafe/atomic code, wrong comments are not cosmetic.

Non-blocking: the new 5s spin/concurrency tests are useful as smoke/stress tests, but they should not be presented as proof of the memory-ordering argument. A loom/model test would be a good follow-up if this code is staying subtle.

orthur2:
> I'm reviewing the comments and order. From agent review, it says:
> 
> Blockers:
> 
> 1. Please explain the Recv::drop interleaving around the RECEIVING -> EMPTY -> DISCONNECTED path. In particular: if a sender concurrently succeeds in send during the temporary EMPTY window, who owns dropping the message, the waker, and the allocation, and in what order? The current code may rely on the next loop iteration doing EMPTY -> DISCONNECTED, but that safety argument is not written down clearly enough for this kind of code.
> 2. Fix the misleading comments in oneshot/mod.rs, especially the poll MESSAGE branch saying the sender was dropped when it actually sent a message, and the stale memory-barrier comment around RECEIVING -> EMPTY. In unsafe/atomic code, wrong comments are not cosmetic.
> 
> Non-blocking: the new 5s spin/concurrency tests are useful as smoke/stress tests, but they should not be presented as proof of the memory-ordering argument. A loom/model test would be a good follow-up if this code is staying subtle.

The agent review makes sense. I focused on the state-machine change and missed that some existing comments became stale or misleading.

I added a small follow-up to fix the MESSAGE / RECEIVING -> EMPTY comments and spell out the Recv::drop case a bit more.

For the temporary EMPTY window, after Recv::drop successfully CAS RECEIVING -> EMPTY with Acquire, it has retracted the published waker and owns dropping that waker. The Acquire CAS synchronizes with write_waker's Release CAS, so the waker is initialized before drop_waker().If a sender's send transition wins during this temporary EMPTY window, it observes EMPTY, publishes MESSAGE, and returns Ok. Because it did not observe RECEIVING, it never takes or drops the waker, and because send succeeded from EMPTY, it does not own deallocating the channel. Recv::drop still owns the retracted waker, drops it, and loops. Once the sender has published MESSAGE, Recv::drop will observe MESSAGE with Acquire, drop the message, and deallocate the channel.If no sender wins that temporary EMPTY window, Recv::drop's next loop iteration can complete EMPTY -> DISCONNECTED. In that case, any later cleanup is owned by the sender side: either Sender::send observes DISCONNECTED and the SendError path owns the message/allocation cleanup, or Sender::drop observes DISCONNECTED and deallocates the channel.

tisonkun:
Thanks! Pending for merging ...

PR #114, opened 2026-06-06
feat: Add KeyedFairSemaphore

tisonkun:
## Summary
- add mea::keyed_fair_semaphore::KeyedFairSemaphore with borrowed and owned permits
- schedule queued waiters by the fewest in-flight permits for their key, preserving queue-order ties
- cover keyed fairness, per-key FIFO, cancellation, capacity stress, docs, and Send/Sync/Unpin assertions

## Verification
- cargo x test
- cargo +nightly x lint


tisonkun:
The implementation is OK. But we shall find a better name for this primitive.

PR #139, opened 2026-08-05
docs(semaphore): fix OwnedSemaphorePermit::forget doctest example

xleoken:
The original doctest for OwnedSemaphorePermit::forget used try_acquire, which returns a borrowed SemaphorePermit instead of OwnedSemaphorePermit, causing a type mismatch.

### DIFF
--- a/mea/src/semaphore/mod.rs
+++ b/mea/src/semaphore/mod.rs
@@ -555,7 +555,7 @@ impl OwnedSemaphorePermit {
     ///
     /// let sem = Arc::new(Semaphore::new(10));
     /// {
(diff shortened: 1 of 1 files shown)


tisonkun (approved):
Thanks! But what do our test coverage doesn't handle this item?

xleoken:
> Thanks! But what do our test coverage doesn't handle this item?

i checked the tests, doesn't cover.

tisonkun:
@xleoken Ah. I got it. The doctest runs but the case is inaccurate.

Thanks for your contribution again!

PR #151, opened 2026-08-13
chore: tidy up meterials post rebrand

tisonkun:


mingcheng:
The CHANGELOG history file doesn't need to be renamed, because this is a continuation of the project.

tisonkun:
> this is a continuation of the project

This is one possible choice. But I choose to archive old changelog, which is also a typically option.

For example:

* https://github.com/apache/arrow-rs/blob/main/CHANGELOG.md
* https://github.com/apache/arrow-rs/blob/main/CHANGELOG-old.md

I always consider contributors can adopt the style they like, so would not spend time on bikeshedding this.

tisonkun:
That is, I'd prefer to archive old changelogs from time to time to keep the most interesting ones short enough to open and search.

Typically, users migrate from recent versions, and keeping an arbitrarily long changelog doesn't help.

I may do the same action for a 1.0 version to archive all pre-1.0 changelog.

Issue #160, opened 2026-08-20
Async version of LazyLock


tisonkun:
I remember I don't provide `LazyLock` before because `Deref`'s method is not async.

But perhaps I can try to figure out an async `LazyLock` with a named async getter but not implement `Deref`.

_Originally posted by @tisonkun in https://github.com/apache/opendal/pull/8108#discussion_r3821385167_

mhambre:
So there's an existing library that has this core idea implemented but it's older so it may either be complete or dead:

Here's an example:
```rust
use async_once_cell::Lazy;
use std::future::Future;
use std::pin::Pin;

type BoxFut<T> = Pin<Box<dyn Future<Output = T> + Send>>;

struct Service {
    client: Lazy<DbClient, BoxFut<DbClient>>,
}

impl Service {
    fn new(url: String, creds: Credentials) -> Self {
        Self {
            client: Lazy::new(Box::pin(async move {
                // `url` and `creds` are moved into the future
                DbClient::connect(&url, creds).await.expect("connect failed")
            })),
        }
    }

    async fn client(&self) -> &DbClient {
        self.client.get_unpin().await
    }
}
```

We might be able to implement this similarly. In reference to usage in OpenDAL the only problem would be that for the issues this came about because of this we'd need to be able to construct an owned instance of the `OperatorContext` (e.g. for the HTTP transporter).

mhambre:
Thinking about it a bit more, we should definitely make sure to include try semantics for `Error` types since the major use case for this would be IO-based futures susceptible to blips that should be retriable at the user's discretion.

I can try taking a swing at it.

tisonkun:
@mhambre Brilliant! Just go ahead :D

tisonkun:
Firts version has been landed via #168.

PR #168, opened 2026-08-21
feat(once): Add async LazyCell implementation

mhambre:
Adds support for an asynchronous `LazyCell` inspired by [async-lazy](https://docs.rs/async-lazy/latest/async_lazy) and the APIs provided by `std::sync::LazyLock` and `std::cell::LazyCell`.

The main addition to this is fallible initialization through `LazyCell::try_force` and being runtime agnostic (async-lazy is married to Tokio through sync primitives) by using asyncband's `Mutex` and `OnceCell`.

Some design choices:
- Cancelled futures resume with the next caller rather than restarting the initialization. Restarts can happen via user-defined fallible errors (e.g. a network blip causing a connection to fail). Users get to choose between retryable failures (monadic error returns) and poisoning on failure (initializer panics) through choosing standard methods versus `try_` methods.
    - Retry state management is serialized through the internal `Mutex`. The state stores at most one pinned attempt, which remains available for the next caller after cancellation. An `Err` removes the completed attempt but retains the `FnMut` initializer, allowing the next queued caller to start its own attempt. A successful value is stored in the `OnceCell`.
    - The caveat to resumability is that all initializers must return a `Send` + `'static` future. `Send` is required because resuming may happen on a different thread, while `'static` is required because the stored future may outlive the caller that started it.
    - For fallible initialization, we opt for `FnMut() -> Future<Output = Result<T, E>>` over `AsyncFnOnce` because returning an error must leave the initializer available for another attempt. This design choice was made particularly because `async` programs open the door to IO-based failures outside of the programmer's control, and we want to allow users to handle those failures gracefully. A good example is the OpenDAL implementation, where the HuggingFace API going down temporarily should not permanently prevent canonicalization.
- Panics in the initializer poison the cell no matter what route the user chooses for initialization.
- Keeping with the APIs of `std::sync::LazyLock` and `std::cell::LazyCell`, `get` methods are introspective, while `force` methods actively perform initialization. We also maintain the use of associated functions over methods.
- Deref traits aren't implemented because you can't await a deref, so all interactions with the `LazyCell` happen through associated functions.
- To support arguments supplied to the initializer, we expose `_with` functions that take a single argument. This can be a tuple or struct to allow for multiple arguments. Because the returned future is `'static`, the initializer can borrow a large argument, clone only the fields it needs, and move those fields into the future:

```rust
let repo = LazyCell::new(|ctx: &OperationContext| {
    let client = ctx.client.clone();
    let token = ctx.token.clone();

    async move { canonicalize(client, token).await }
});

let repo = LazyCell::try_force_with(&repo, &ctx).await?;
```

This implementation is in reference to #160. An example of it used in practice inside of OpenDAL's HuggingFace service can be found [here](https://github.com/mhambre/opendal/tree/chore/move-to-lazylock).


tisonkun (approved):
Thanks for nudging and demonstrating this feature @mhambre!

PR #188, opened 2026-08-25
perf(semaphore): Allocate wake buffer only when needed

mhambre:
Avoids unnecessary wake buffer allocation on the semaphore release hot path when no tasks need to be woken.

```bash
cargo bench -p benchmarks --bench benchmarks -- semaphore mutex \
  --min-time 0.75 \
  --sample-count 100 \
  --color never
```

## Median

| Benchmark | Before | After | Improvement |
|---|---:|---:|---:|
| Mutex uncontended reuse | 57.94 ns | 21.15 ns | **63.5%** |
| Semaphore try-acquire/release | 51.10 ns | 13.82 ns | **73.0%** |
| Semaphore owned acquire/release | 52.40 ns | 22.61 ns | **56.9%** |
| Semaphore release | 90.83 ns | 43.93 ns | **51.6%** |
| Repeated permit debt | 4.124 µs | 1.728 µs | **58.1%** |

## Mean

| Benchmark | Before | After | Improvement |
|---|---:|---:|---:|
| Mutex uncontended reuse | 57.99 ns | 21.24 ns | **63.4%** |
| Semaphore try-acquire/release | 51.00 ns | 13.85 ns | **72.8%** |
| Semaphore owned acquire/release | 54.46 ns | 22.62 ns | **58.5%** |
| Semaphore release | 94.47 ns | 43.62 ns | **53.8%** |
| Repeated permit debt | 4.124 µs | 1.731 µs | **58.0%** |


tisonkun (approved):
Thanks!

PR #198, opened 2026-08-26
feat(lazy-cell): add explicit constructors

### LINKED ISSUE #194: Add explicit LazyCell constructors for futures and values
## Summary

Add explicit constructors for the two non-closure states supported by `LazyCell`:

```rust
let pending = LazyCell::from_future(async { 42 });
let ready = LazyCell::from_value(42);
```

Keep the constructor meanings distinct:

- `LazyCell::new` stores an `FnOnce() -> Fut`, so even creation of the future is deferred until the cell is forced.
- `LazyCell::from_future` stores an already-created future. Construction and pinning may happen immediately, but the future must not be polled until the cell is forced.
- `LazyCell::from_value` creates an already-initialized cell and should be a `const fn` defined on `impl<T> LazyCell<T>`, fixing the initializer parameter to its default type and matching `OnceCell::from_value`.
- `LazyCell::default` remains lazy: it initializes `T::default()` when forced.

Remove the existing `impl<T, F> From<T> for LazyCell<T, F>`. The named `from_value` constructor is explicit, const-capable, and inferable without a target type. In contrast, `LazyCell::from(value)` cannot infer `F` without a type annotation and generates `Into<LazyCell<T, F>>` conversions for every initializer type even though an initialized cell contains no initializer.

Do not add `From<Fut>` after removing `From<T>`. Although removing the value conversion eliminates their direct coherence conflict, an explicit `from_future` method makes the construction-versus-polling timing distinction discoverable and avoids reopening the generic conversion surface.

`from_future` should accept `Future<Output = T> + Send + 'static`, preserve the same in-flight future across caller cancellation, and return the ordinary `LazyCell<T>` type. Add rustdoc and tests covering these observable behaviors:

- `from_value` is immediately visible through `get` and works in a const context without a target type annotation.
- `from_future` does not poll its future before `force`.
- Cancellation of a `force` caller preserves and resumes the same future instead of recreating it.
- No `From<T>` or `From<Fut>` conversion remains for `LazyCell`.

pollychen-lab:
## Summary
Adds explicit `LazyCell::from_future` and const `LazyCell::from_value` constructors, removes the ambiguous `From<T>` conversion, and adds coverage for lazy polling, cancellation resumption, and const initialization.

Closes #194

### DIFF
--- a/asyncband/src/once/lazy_cell/mod.rs
+++ b/asyncband/src/once/lazy_cell/mod.rs
@@ -215,6 +215,38 @@ impl<T, F> LazyCell<T, F> {
     }
 }
 
+impl<T> LazyCell<T> {
+    /// Creates a new `LazyCell` from an already-created asynchronous future.
+    ///
+    /// The future is stored without being polled. Calling [`force`](Self::force) starts polling
+    /// it, and cancellation preserves the in-flight future for a later caller.
+    pub fn from_future<Fut>(future: Fut) -> Self
+    where
+        Fut: Future<Output = T> + Send + 'static,
+    {
+        Self {
+            value: ValueCell::new(),
+            state: Mutex::new(State {
+                initializer: None,
+                attempt: Some(Box::pin(future)),
+            }),
+            poisoned: AtomicBool::new(false),
+        }
+    }
+
+    /// Creates a new `LazyCell` that already contains `value`.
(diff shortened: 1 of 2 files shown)


tisonkun (approved):
Thanks for your contribution @pollychen-lab! LGTM.

Merging ...

PR #219, opened 2026-08-26
test(channel): add ecosystem performance baselines

### LINKED ISSUE #208: test(channel): establish ecosystem performance baselines
Parent: #206

## Goal

Define repeatable performance baselines for channel work against semantically similar Rust implementations.

## Scope

Cover ready paths and concurrent bounded and unbounded workloads for these shapes:

- one producer and one consumer;
- many producers and one consumer;
- one producer and many competing consumers;
- balanced many-producer, many-consumer contention;
- broadcast fanout and producer contention when broadcast exists;
- latest-state notification when watch exists.

Use comparable semantics and document exceptions such as Tokio broadcast's overwrite and lag behavior. Measure multiple participant counts so wake-all or contention cliffs are visible.

## Acceptance criteria

- The benchmark methodology and peer versions are documented.
- Results are reproducible locally through the repository workflow or a documented companion workflow.
- Sustained regressions above 3x are investigated, and order-of-magnitude regressions block an implementation from being accepted unless the semantic tradeoff is explicit.
- Third-party benchmark dependencies do not become runtime dependencies of `asyncband`.

tisonkun:
## Summary

- add a standalone ecosystem benchmark target for the current bounded and unbounded MPSC APIs
- compare ready round trips and 1/2/4/8-producer contention with Tokio, async-channel, and flume
- organize Asyncband self-benchmarks and ecosystem comparisons as sibling suites, each grouped by primitive and operation
- keep all peer implementations confined to benchmark dev dependencies

Closes #208.

## Design Notes

This PR was stacked on #217; GitHub retargeted it to `main` after #217 merged.

The benchmark crate root contains only `Cargo.toml`. Asyncband self-benchmarks live under `benchmarks/asyncband/<primitive>/<operation>.rs`, while cross-project comparisons live under `benchmarks/ecosystem/<primitive>/<operation>.rs`. Shared setup stays in the nearest `support.rs`, and the existing self-benchmark scenarios remain unchanged.

Tokio is the closest MPSC peer. async-channel and flume expose MPMC receivers, so the suite measures them with one receiver and records that topology difference. The compared versions are Tokio 1.53.1, async-channel 2.5.0, and flume 0.12.0; `Cargo.lock` records the exact resolution.

Bounded channels use capacity 64. Concurrent cases move 16,384 `usize` messages from 1, 2, 4, or 8 producer threads to one consumer. Channel construction and worker lifecycle stay outside the timed section, and async paths are driven without installing a peer-specific runtime. Run the full workflow with `cargo x bench`, or focus this target with `cargo bench -p benchmarks --bench ecosystem`.

These comparisons are regression signals rather than a fastest-wins contest. A sustained result above 3x the closest semantic peer warrants investigation; an order-of-magnitude gap should block the implementation unless a documented semantic or resource tradeoff explains it.


benchmarks/README.md
tisonkun, line 1: This is to-be-removed.

Issue #206, opened 2026-08-26
Tracking Issue to deliver the channel family incrementally


tisonkun:
## Goal

Deliver the channel taxonomy agreed in #167 through small, independently reviewable pull requests instead of one bundled implementation.

## Current status for 0.7

- The source tree and public modules now share direct crate-root families; #234 removed the intermediate private `channel` source wrapper without changing public paths.
- The ecosystem benchmark baseline is complete in #208.
- Lossless unbounded MPMC broadcast is implemented at `asyncband::broadcast::mpmc::unbounded` by #117; #213 now tracks only the bounded variant.
- The latest-state `watch` channel is complete in #215 and #233.
- The Cargo feature decision is complete: `oneshot`, `mpsc`, `broadcast`, and `watch` remain independent opt-in features, with no `channel` umbrella for 0.7. See #216 and #226.

## Working agreements

- Keep public modules and their source families at direct crate-root paths such as `asyncband::mpsc` and `asyncband/src/mpsc`.
- Put machinery shared by multiple families under the private crate-level `internal` module; keep family-specific machinery inside that family.
- Preserve existing public behavior unless a subtask explicitly proposes an API change.
- Add one channel family or one implementation concern at a time.
- Compare contention and topology-specific performance with semantically similar Rust channels. Being the fastest is not required, but sustained order-of-magnitude regressions must be treated as implementation defects.
- Keep bounded backpressure and unbounded growth as the initial policies. Sliding, drop-oldest, and lag semantics remain deferred.
- Do not expose queue, ring, Disruptor, or sequencer machinery through the public API.

## Remaining delivery

- #209: reconsider the existing MPSC backend.
- #210: decide whether an SPSC queue is justified.
- #211: add a competing MPMC queue.
- #212: add a competing SPMC queue after the MPMC design.
- #213: add bounded lossless MPMC broadcast.
- #214: evaluate an SPMC broadcast specialization.

Design baseline: #167. Prior broadcast work: #117 and #146.

tisonkun:
The first implementation step is open as #217 and covers #207 only. It preserves the existing oneshot and MPSC algorithms, public paths, behavior, and Cargo features.

tisonkun:
One follow-up API consideration for bounded queues is **public capacity reservation**, distinct from the private sequence/publication reservations discussed in #167.

[...]

Issue #167, opened 2026-08-21
design(channel): organize channel families and capacity contracts


tisonkun:
## Decision summary

The public API is organized by delivery semantics and keeps common paths at the crate root. There is no public `asyncband::channel` or `asyncband::queue` namespace.

The source tree mirrors the public families directly under `asyncband/src/`. Machinery shared by multiple families belongs under the private crate-level `internal` module, without adding a public path segment or restoring a private `channel` wrapper.

For 0.7, the implemented families are `oneshot`, bounded and unbounded `mpsc`, `broadcast::mpmc::unbounded`, and `watch`. The remaining taxonomy nodes are additive directions rather than release commitments.

The taxonomy has three independent axes:

1. the delivery family: oneshot, competing queue, broadcast, or watch;
2. endpoint topology when it changes endpoint capabilities or enables a static specialization;
3. capacity or retention selected by a constructor.

The taxonomy does not require every node to be implemented in 0.7. Only implemented modules are exported. The reserved sibling modules can be added later without changing existing paths or endpoint contracts.

## Public taxonomy

```text
asyncband
├── oneshot
│   └── channel<T>() -> (Sender<T>, Receiver<T>)
├── watch
│   └── channel<T>(initial: T) -> (Sender<T>, Receiver<T>)
├── mpsc                         # competing queue; implemented family
│   ├── bounded<T>(capacity: usize)
│   └── unbounded<T>()
├── spsc                         # planned competing-queue family
│   ├── bounded<T>(capacity: usize)
│   └── unbounded<T>()
├── spmc                         # planned competing-queue family
│   ├── bounded<T>(capacity: usize)
│   └── unbounded<T>()
├── mpmc                         # planned competing-queue family
│   ├── bounded<T>(capacity: usize)
│   └── unbounded<T>()
└── broadcast
    ├── spmc                     # single producer, multiple subscribers; planned
    │   ├── bounded<T>(capacity: usize)
    │   └── unbounded<T>()
    └── mpmc                     # multiple producers, multiple subscribers
        ├── bounded<T>(capacity: usize)
        └── unbounded<T>()
```

Every bounded constructor returns `BoundedSender<T>` and `BoundedReceiver<T>` from its leaf module. Every unbounded constructor returns `UnboundedSender<T>` and `UnboundedReceiver<T>`.

The common queue path remains short:

```rust
use asyncband::mpsc;

let (tx, mut rx) = mpsc::bounded(128);
tx.send(event).await?;
[...]

PR #226, opened 2026-08-27
docs(channel): clarify feature boundaries

tisonkun:
## Summary

* Keep the existing `mpsc` Cargo feature instead of introducing a speculative `queue`, `dispatch`, or `channel` aggregate.
* Document that current channel features follow current public modules: `oneshot`, `mpsc`, and `broadcast`; topology and capacity remain API-level choices.
* Separate one-shot, single-delivery MPSC, and broadcast entries in the API guide.
* Correct the changelog wording: `broadcast::overflow` was removed, while the `broadcast` feature remains.

## Design Notes

Cargo features control compiled capabilities and dependencies; they do not need to encode every channel topology. The `broadcast` feature can expose specialized SPMC and MPMC APIs under one delivery family, while implementation-specific producer topology remains visible in module paths and types. Additional aggregate or leaf features should be introduced only when implemented variants demonstrate a material compile-time, dependency, binary-size, or audit boundary.

Refs #167 and #206.

Validation: `cargo x check`, `cargo x build`, `cargo x lint`, `cargo x test`, and `cargo +1.86.0 check -p asyncband --no-default-features --features mpsc`.


asyncband/src/lib.rs
```diff
@@ -49,17 +49,24 @@
 //! # }
 //! ```
 //!
+//! Channel features currently match public API modules rather than speculative aggregate families:
+//! `oneshot`, `mpsc`, and `broadcast`. Nested variants such as [`broadcast::mpmc::unbounded`]
+//! express topology and capacity in module paths and constructor names without additional Cargo
+//! features.
```
tisonkun:
```suggestion
```

README.md
```diff
@@ -56,6 +56,8 @@ async fn increment() {
 
 Public paths stay direct—such as `asyncband::mutex`, `asyncband::pool`, and `asyncband::once::OnceCell`—while Cargo features keep unused implementations out of the build.
 
+Channel features currently match public API modules rather than speculative aggregate families: `oneshot`, `mpsc`, and `broadcast`. Nested variants such as `broadcast::mpmc::unbounded` express topology and capacity in module paths and constructor names without additional Cargo features.
+
```
tisonkun:
```suggestion

```

PR #233, opened 2026-08-28
feat(watch): add a latest-state channel

### LINKED ISSUE #215: feat(watch): add a latest-state channel
Parent: #206

## Goal

Add a latest-state watch channel in which receivers independently track whether they have observed the current version while intermediate updates may coalesce.

## Contracts

- The channel always retains one current value.
- Each receiver tracks its own observed version.
- `borrow` does not implicitly consume a change unless the selected API explicitly says so.
- A final unseen value remains observable before disconnection is reported.

## Acceptance criteria

- Cover receiver cloning, subscriptions, change coalescing, cancellation, final-value draining, disconnection, and dropping replaced values outside internal locks.
- Compare notification and borrow/update paths with Tokio watch.
- Keep versioning and waiter machinery private.

Depends on #208 and the taxonomy in #167.

QwQBiG:
## Summary

- add an opt-in `asyncband::watch` latest-state channel with cloneable senders and independently versioned receivers
- retain one current value, coalesce intermediate updates, keep `borrow` non-consuming, and drain a final unseen value before reporting disconnection
- make `changed` cancellation safe and release replaced values, retired wakers, and wake callbacks outside the internal lock
- add contract and trait tests, public documentation, self-benchmarks, and Tokio 1.53.1 ecosystem comparisons

Closes #215.

## Design Notes

The public API returns owning `Arc<T>` snapshots from `borrow`, `borrow_and_update`, and `changed`. Unlike Tokio's lock-backed `Ref<T>` and `changed() -> Result<(), _>` API, these snapshots never retain the channel's internal lock and may be stored or moved independently. The ecosystem benchmark documents this semantic difference and immediately reads the current `usize` value on both sides.

The implementation retains an `Arc<T>`, a monotonically checked version, endpoint counts, and a `WaitSet` under one private mutex. Each receiver stores its own observed version. New subscriptions begin at the current version, while receiver clones inherit the source receiver's observed version and then advance independently.

The internal `arena`, `mutex`, and `waitset` feature gates include `watch` because of the `watch -> WaitSet -> Arena` dependency chain and watch's direct use of the internal mutex. These additions are additive and do not change any existing feature combination or public behavior.

`WaitSet::is_empty()` lets the watch send and final-sender-drop paths avoid taking an empty waker set. Besides avoiding unnecessary work, this skips the internal epoch advance performed by `take_wakers()` when no waiter exists.

The benchmarks cover current-value borrowing, send-and-borrow paths, ready changes, pending notification, and 1/2/4/8/32-receiver fanout. No measured watch path showed a sustained regression above 3x relative to Tokio on the development machine.

Validation included `cargo x build --locked`, `cargo x check`, `cargo x test --no-capture`, `cargo x bench`, Rust 1.86 watch-only all-target checks, nightly Clippy and rustfmt, typos, Hawkeye license checks, and all-features rustdoc with warnings denied.


tisonkun (approved):
Thanks for your contribution @QwQBiG!

I'll push some follow-up to polish code and APIs but this one is good enough to merge AS IS.

PR #240, opened 2026-08-29
feat(completion): add one-shot completion primitive with shared receivers

### LINKED ISSUE #222: Provide a shared one-shot completion primitive
## Background

Asyncband currently provides two nearby but different contracts:

- `oneshot` transfers one value from one sender to one receiver;
- `OnceCell` stores one initialized value, with initialization driven through the cell API.

Neither directly models one producer completing a result that can be observed by multiple independent waiters, including waiters created after completion. This pattern is known as `Deferred` in [Cats Effect](https://typelevel.org/cats-effect/docs/std/deferred), `Promise` in [ZIO](https://zio.dev/reference/concurrency/promise/), and `CompletableDeferred` in [Kotlin coroutines](https://kotlinlang.org/api/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines/-completable-deferred/).

The primitive is useful when one operation publishes a handshake result, negotiated metadata, configuration, or initialization outcome to several dependent operations. It does not run that operation and does not spawn tasks; it only stores and distributes its eventual result.

Tracked by #218.

## Proposed minimum contract

The API should separate producer and observer capabilities, provisionally called `Completer<T>` and `Completion<T>`:

- there is one completion capability and a cloneable observer capability;
- the value can be completed exactly once;
- completion wakes all current observers;
- observers created or awaited after completion immediately see the same value;
- cancelling one wait only unsubscribes that waiter;
- dropping the completer without a value transitions observers to an explicit closed state rather than leaving them pending forever;
- the primitive remains runtime agnostic and makes progress only through explicit calls, drop, and polling.

The final module and type names remain an API decision. `Completion` describes the mechanism without importing JavaScript's full `Promise` scheduling and chaining expectations.

### Shared value representation

A value cannot be moved independently into multiple observers without an ownership policy. The main options are:

1. store `Arc<T>` internally and return `Arc<T>`, avoiding a `Clone` bound but exposing shared ownership;
2. return a borrow tied to the observer handle, which is allocation-efficient but complicates owned futures and lifetimes;
3. require `T: Clone` when awaiting, which keeps returned values ordinary but excludes non-cloneable results and makes wake cost depend on observer count; or
4. return a dedicated guard that dereferences to `T`, hiding but not eliminating shared storage.

This choice is observable and must be settled before implementation. Requiring `T: Clone` on the primitive itself should be avoided unless a concrete API benefit justifies it.

### Producer cardinality

A single non-cloneable completer makes the producer role clear and gives completer drop an unambiguous closed transition. A cloneable completer would support racing producers, but then closure could occur only after every producer is dropped and the API would overlap more heavily with `OnceCell::set`.

The preferred initial direction is one non-cloneable completer. `complete(value)` should report failure and return ownership of `value` if another completion has already won or no observers remain, rather than silently discarding it.

### Completion payload versus transport error

The primitive should store an arbitrary `T`. Callers can choose `T = Result<V, E>` when completion itself represents success or failure. Closure caused by dropping the completer is transport state and must remain distinct from a user-provided error value.

### Observer lifecycle

`Completion<T>` should be cloneable. Dropping one observer must not affect other observers or erase an already completed value. The issue should decide whether dropping all observers lets the completer detect disconnection and recover a submitted value, mirroring channel-style producer feedback, or whether completion remains valid without current observers.

### Cancellation and waiter cleanup

A pending wait is cancel safe: dropping it removes or invalidates only its waiter registration. Completion racing with cancellation must result in either that waiter receiving the shared result or being cleanly removed, with no lost wake-up and no stale waker retained indefinitely.

### Relationship with existing primitives

`OnceCell` may provide useful storage concepts, but shared completion adds a wait-only capability, explicit producer closure, and multi-observer wake-up semantics. `oneshot` may provide endpoint and error naming patterns, but changing oneshot to clone its receiver would alter its ownership contract. The new primitive should therefore remain nominally distinct even if it reuses internal machinery.

## Key trade-offs

- `Arc<T>` or a guard versus borrowed output versus `T: Clone`.
- A single completer with simple closure semantics versus multiple racing completers.
- Channel-style disconnected feedback versus completion that remains valid with zero current observers.
- Reusing `OnceCell` or oneshot internals versus a small dedicated state machine.
- Returning a named wait future versus an `async fn`, especially for owned and cancellation-safe variants.

## Non-goals

The first version should not include:

- task spawning or execution ownership;
- promise chaining, callbacks, executors, or microtask scheduling;
- reset or reuse after completion;
- progress updates or multiple values;
- built-in timeout, retry, or cancellation of the producer's work;
- implicit error aggregation.

## Acceptance criteria

- Producer and observer capabilities are distinct and documented.
- Exactly one value can win completion, and losing completion attempts preserve ownership of their value.
- Every current and future observer receives the same completed result under the selected ownership model.
- Dropping the completer before completion produces a documented closed result.
- Cancelling or dropping one observer does not affect other observers and does not leak its waker.
- Races among completion, completer drop, observer drop, and waiter cancellation have deterministic regression tests.
- The distinction from `oneshot` and `OnceCell` is documented with examples.
- No executor, timer, or runtime-specific dependency is introduced.
- Repository build, test, lint, and formatting workflows pass through `cargo x`.

QwQBiG:
## Summary

- add an opt-in, runtime-agnostic `asyncband::completion` primitive with one non-cloneable `Completer` and cloneable `Completion` observers
- publish one value exactly once, wake all current waiters, retain it for late observers, and report an explicit closed state when the completer is dropped before completion
- return ownership of rejected values after duplicate completion or when no observers remain
- add public documentation, a shared-result example, contract and trait tests, race and cancellation regressions, and self-benchmarks

Closes #222.

## Design Notes

`Completion::wait(&self) -> Result<&T, WaitError>` returns a borrow tied to the observer instead of imposing `T: Clone` or always exposing `Arc<T>`. This keeps the primitive's ownership policy minimal: callers can borrow the stored value directly, clone it when appropriate, or choose `T = Arc<U>` when independently owned shared results are needed.

The primitive remains distinct from both nearby APIs. `oneshot` transfers one value to one receiver, while completion retains one result for multiple current and future observers. `OnceCell` drives initialization through the cell API, while completion separates a producer-only capability from wait-only observer capabilities and reports producer closure explicitly. The `shared_completion` example demonstrates one worker publishing a result to two independent consumers and a late observer; expressing the same lifecycle with oneshot would require one channel per consumer, manual fan-out, and separate storage for late consumers.

Internally, `Completion` observers own the strong `Arc<Shared<T>>`, while `Completer` holds a `Weak<Shared<T>>`. This prevents the completer from retaining the shared state or a completed value after the final observer is gone. `complete()` temporarily upgrades the weak reference, and the mutex-protected observer count remains the disconnection authority so completion and final-observer drop linearize cleanly. The completed value is stored in `OnceLock<T>`, while the same small private mutex protects the `Pending | Completed | Closed` status, observer count, and `WaitSet`. `WaitError::Closed` specifically means that the completer was dropped without producing a value; it remains distinct from any error carried by the user-provided payload itself, such as `T = Result<V, E>`. Each `wait()` owns an independent `WakerToken`, so cancellation unregisters only that call. Waker replacement, cancellation cleanup, wake callbacks, and rejected payload destruction occur after releasing the internal lock. The existing `wake_all` behavior ensures that every remaining waiter is attempted even if one waker panics.

The internal module `#[cfg]` gates for `arena`, `mutex`, and `waitset` now include `feature = "completion"`: completion uses the internal mutex directly and reaches `Arena` through `WaitSet`. These additions only make existing private implementation modules available when the public `completion` feature is enabled; they do not alter any other public feature combination or public behavior.

The benchmarks cover ready waits, pending cancellation, complete-then-wait, pending notification, and 1/2/4/8/32-observer fanout. They are self-benchmarks because the existing primitives do not provide a semantically equivalent retained multi-observer result.

Validation included `cargo x build --locked`, the nightly feature-matrix `cargo x check`, full workspace tests on stable and Rust 1.86.0, `cargo x bench`, nightly Clippy and rustfmt with warnings denied, Taplo checks for every changed manifest, typos, Hawkeye license checks, and all-features rustdoc with warnings denied.


asyncband/src/completion/mod.rs
```diff
@@ -0,0 +1,262 @@
...
+        f.debug_struct("Completer").finish_non_exhaustive()
+    }
+}
+
+impl<T> Completer<T> {
+    /// Completes the primitive with `value` and wakes all pending observers.
+    ///
+    /// The value is rejected and returned if another value has already completed the primitive or
+    /// if no observers remain.
+    pub fn complete(&self, value: T) -> Result<(), CompleteError<T>> {
```
orthur2: Could we document the panic behavior here? A waker can panic after the value has been committed, so complete() may panic even though completion succeeded.

asyncband/src/completion/mod.rs
```diff
@@ -0,0 +1,262 @@
...
+    Completed,
+    Closed,
+}
+
+/// The capability that completes a [`Completion`] with one value.
+///
+/// This type deliberately does not implement [`Clone`]. Dropping it before completion closes the
+/// primitive and wakes all pending observers.
+pub struct Completer<T> {
+    shared: Arc<Shared<T>>,
```
orthur2: Could `Completer` hold a `Weak<Shared<T>>` instead?

Once the last `Completion` is dropped, there is no way to create another observer, but the current strong `Arc` can still keep a completed value alive until the completer itself is dropped. Since the API already treats having no observers as disconnected, it seems more natural for the result to be released with the last `Completion`.

orthur2 (commented):
Nice work, and thanks for the contribution! I left one question and one small docs note inline.

asyncband/src/completion/mod.rs
QwQBiG: Good catch — I've added a `# Panics` section documenting that the value is committed before notification begins, so a waker panic can occur after completion succeeds. It also notes that `complete()` still attempts to wake the remaining registered wakers before resuming the panic.

asyncband/src/completion/mod.rs
QwQBiG: Good point — I've switched `Completer` to `Weak<Shared<T>>`, while `Completion` observers retain the strong `Arc`. The mutex-protected observer count remains in place to linearize `complete()` against the final observer drop, and I added a regression test confirming that the completed payload is released with the last observer.

tisonkun (approved):
Thanks for your contribution @QwQBiG!

I push a few commits to:

1. Fix the Send/Sync bound to `Completer`
2. Let `Completer::complete` consume `self` to retain the semantic like C++'s promise + shared_future.
3. Naming changes on error handling.

Merging ...

Feel free to drop post merge review if you find anything that can be improved further.

asyncband/src/completion/mod.rs
```diff
@@ -92,29 +85,41 @@ struct Shared<T> {
...
 enum Status {
     Pending,
     Completed,
-    Closed,
+    Abandoned,
+}
+
+/// The error returned by [`Completion::wait`] when the completer was dropped without a value.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub struct Abandoned;
```
tisonkun, line 100: This may be refactored to `pub struct Abandoned(())` to avoid outside construction?

QwQBiG: Ah, I missed the existing `PoolIsEmpty(())` precedent over in the pool module. Using the same private-field pattern for `Abandoned` makes sense too. I’ll give the merged code another pass — that one slipped past me, my bad qwq

QwQBiG: > This may be refactored to `pub struct Abandoned(())` to avoid outside construction?

Yes — this can be changed without affecting the runtime contract. The private `()` field keeps `Abandoned` zero-sized and `Copy`, while preventing downstream code from constructing or destructuring it, so the error is only produced by the completion API. It also matches the existing `PoolIsEmpty(())` pattern. Right?

The only source-level fallout is that the integration tests which currently construct `completion::Abandoned` directly will need to use `matches!` or inspect the returned error instead

If you want…I can send a tiny follow-up PR for this. qwq

tisonkun: Feel free to send one :D

PR #248, opened 2026-08-30
refactor: harden library ergonomics and test contracts

tisonkun:
## Summary

- Remove latch future types that callers could neither construct nor obtain through the public async methods.
- Reject zero-sized bounded pools instead of allowing checkouts that can never complete.
- Replace debug-string assertions, scheduler sleeps, and pseudo pressure benchmarks with deterministic tests of observable behavior.
- Clarify lifecycle, cancellation, capacity, and ownership contracts across the public documentation, and correct misleading examples.

## Design Notes

The audit treats Asyncband as a foundational published crate: public surface area must be usable, examples must teach sound async patterns, and tests should defend stable behavior rather than formatting or scheduler timing. Each concern is isolated in a small commit so it can be reviewed or dropped independently.

The patch removes 1,190 net lines. The only runtime behavior change is rejecting a bounded pool whose maximum size is zero; the rest narrows unusable API surface, strengthens the default-feature test workflow, or improves tests and documentation.

## Validation

- `cargo x check`
- `cargo x test`
- `cargo x lint`
- `cargo package -p asyncband --allow-dirty`


asyncband/src/shutdown/mod.rs
```diff
@@ -180,10 +185,7 @@ impl ShutdownGuard {
         self.latch.wait().await;
     }
 
-    /// Returns an owned future that resolves when shutdown is requested.
-    ///
-    /// The returned future has no lifetime constraints and does not keep shutdown completion
-    /// pending.
+    /// Returns a future that can be spawned and resolves when shutdown is requested.
```
tisonkun:
```suggestion
    /// Returns a future that resolves when shutdown is requested.
    ///
    /// The returned future can be moved into a spawned task.
```

PR #249, opened 2026-08-30
refactor(completion): make Abandoned opaque

QwQBiG:
## Summary

- give `completion::Abandoned` a private unit field, following the existing `PoolIsEmpty(())` pattern, so downstream code cannot directly construct or destructure it
- update integration tests to verify abandonment without constructing the error
- preserve its zero-sized layout, `Copy` semantics, and completion behavior

Follow-up to #240.


tisonkun (approved):
Thanks!

Issue #220, opened 2026-08-26
Provide a dynamically registered phaser


tisonkun:
## Background

Asyncband currently provides several related coordination primitives, but none covers reusable phases with a participant set that changes over time:

- `Barrier` is reusable but fixes the participant count at construction time.
- `Latch` is a one-way countdown.
- `WaitGroup` tracks a dynamic set of handles, but it represents one completion epoch rather than repeated rounds.

A phaser fills the gap for iterative work whose participants may join or leave between rounds. Java's [`Phaser`](https://docs.oracle.com/en/java/javase/25/docs/api/java.base/java/util/concurrent/Phaser.html) is the main prior art: parties register dynamically, arrive separately from waiting, optionally deregister on arrival, and observe a monotonically advancing phase.

This primitive fits Asyncband's runtime-independent scope because progress is driven only by explicit registration, arrival, drop, and polling. It does not need to spawn tasks, obtain an executor, or install timers.

Tracked by #218.

## Proposed minimum contract

The first version should provide one non-hierarchical phaser with these capabilities:

- observe the current phase;
- register one participant and optionally register several participants at once;
- record an arrival without waiting;
- arrive and wait for the phase to advance;
- arrive and deregister so the participant is excluded from later phases;
- wait for a particular observed phase to advance without implicitly registering another participant;
- wake every waiter for a completed phase exactly once.

Registration racing with phase advancement must have one documented linearization point: it joins either the current phase or the next phase, never an ambiguous mixture of both.

The primitive must remain runtime agnostic and use the standard `Future`/`Waker` contract.

### Participant representation

Java exposes integer counts and treats arriving without registration as a usage error. Rust can make this safer by returning a participant capability from registration. A participant handle could prevent double arrival in one phase and make deregistration explicit in the type system.

The main alternatives are:

1. counter-oriented methods directly on `Phaser`, which are compact and match Java but permit more misuse; or
2. RAII participant handles, which add types but can encode one registration and its lifecycle.

A participant handle is the preferred initial direction, but the public shape should be settled before implementation.

### Drop and deregistration

A registered participant that disappears without arriving can prevent a phase from advancing forever. Automatically deregistering on handle drop avoids that leak, but an implicit drop can also advance a phase at a surprising point.

The issue must decide whether drop means:

- arrive and deregister from the current phase;
- deregister only from future phases while the current arrival remains required; or
- a usage error requiring explicit deregistration.

[...]

PR #250, opened 2026-08-30
feat(phaser): add reusable phase coordination

ButterBright:
## Summary

- Add an opt-in, runtime-agnostic `Phaser` for repeated rounds with dynamic, owned participants. `register_one()` registers one participant; `register(n)` atomically registers a batch and returns an owning iterator.
- Support participant `wait()`, split `arrive()` / `wait()`, independent `Phaser::wait(observed)` observers, and explicit `close()` with an opaque `Closed` error.
- Add practical examples for shared dictionary construction, dynamic workers, coordinator processing between rounds, and grouped coordination with failure propagation.
- Cover public behavior in integration tests, retain counter-wrap tests internally, and add Phaser benchmarks and Miri coverage. Validated with 521 tests on the current toolchain and Rust 1.86, lint/docs, the feature matrix, workspace builds, and all three runnable examples. Miri passes the supported tests; OS-backed Tokio tests are skipped there.

## Design Notes

A participant owns one arrival obligation per phase. Repeated arrivals in that phase count once. Cancelling a polled participant wait preserves its committed arrival and pending phase, so retrying observes the same round. Dropping a participant or the unconsumed portion of a batch withdraws those registrations; withdrawal does not certify successful application work. Applications can close the phaser on failure.

Empty phasers remain dormant and reusable. Closing freezes the unfinished phase and releases its waiters with `Closed`; already completed phases remain successful. Phase numbers are wrapping `u64` observations: waiting detects a change, not arrival at a future numeric target.

Waker registration uses the existing borrowed-waker API inside one state critical section, following the contract adopted in #257. Repeated polls reuse a matching registration; replacement cleanup and notification run outside the lock. The examples express grouped coordination and asynchronous work between rounds by composing phasers and coordinator tasks.


orthur2 (dismissed):
Could you rebase this onto current main branch first? Since the branch was last updated, #273 replaced `WaitSet` with `WakerSet`, and the recent API-table changes now conflict with this branch.

I'd also recommend adding a short `Summary` to the PR description, in line with `AGENTS.md`.You can refer to the body of PRs that have already been merged.

I'd be happy to do a more thorough review once it's rebased.

CHANGELOG.md
```diff
@@ -45,6 +45,7 @@ This non-ASF release was not approved by the Apache Incubator PMC, is not an act
 * Add opt-in `asyncband::once::LazyCell` for values that own one asynchronous initializer and preserve its in-flight future across caller cancellation.
 * Add opt-in bounded and unbounded runtime-agnostic object pools under `asyncband::pool`.
 * Add an opt-in `asyncband::blocking::FutureExt` bridge with `block_on` and `wait_timeout` methods for waiting on runtime-agnostic futures from synchronous code.
+* Add an opt-in runtime-agnostic `Phaser` with dynamic RAII participants, reusable phases, and cancellation-resilient arrival semantics.
```
orthur2: Could you move this entry to Unreleased list?

asyncband/src/phaser/mod.rs
```diff
@@ -0,0 +1,525 @@
...
+//!
+//! # Phase identity
+//!
+//! Phases advance using wrapping arithmetic.
+//!
+//! [`Phase`] supports equality but no ordering or arithmetic contract across wraparound.
+//!
+//! Waiters compare phase identity instead of inferring transitions from party counts.
+//!
+//! # Examples
```
orthur2: Could we add an async example showing a few phases, including separate arrival and waiting?

orthur2: Currently, a spawned task needs a separate `Arc<Phaser>` to call `wait_for_advance()` after `participant.arrive()`. The participant already holds that same `Arc` internally. I'd suggest exposing a `phaser()` accessor so callers can use the two operations separately without carrying another handle.

asyncband/src/phaser/mod.rs
```diff
@@ -0,0 +1,525 @@
...
+//!
+//! Registration increases both the registered and current unarrived counts.
+//!
+//! Arrival reduces the unarrived count.
+//!
+//! Deregistration also removes a party from later phases.
+//!
+//! [`Phaser::arrived_parties`] is the difference between the registered and unarrived counts.
+//!
+//! All state transitions and waiter registration share one synchronization point.
```
orthur2: Could we spell out the memory visibility guarantee here?

The shared synchronization point explains how the internal transitions are ordered, but does not explicitly tell callers whether operations before each participant's arrival happen-before operations after a wait observes that phase's completion. The shared mutex already provides this guarantee, but it should be explicit in the public API docs.

I added a [# Synchronization](https://github.com/apache/asyncband/blob/93b3413dfcde09aaca937848ef7c1e815bbbb79e/asyncband/src/event/mod.rs#L72-L79)  section for this in ManualResetEvent last week. Something similar would help here.

orthur2 (commented):
Thanks for the update! 
I went through the implementation against #220, and it looks good to me. I left a couple of comments on the synchronization guarantee and using the API from a spawned task.

tisonkun (approved):
Thanks for your contribution! I reuse your codebase and add some bench baseline and modify the API a bit.

Welcome to drop a review after merge and if anything can be improved, feel free to submit a new PR.

PR #253, opened 2026-08-31
feat(broadcast): add a bounded MPMC broadcast channel

### LINKED ISSUE #213: feat(broadcast): add a bounded lossless MPMC channel
Parent: #206

## Status

The unbounded half of the original issue is complete. PR #117 implemented `asyncband::broadcast::mpmc::unbounded`, and #234 finalized its source layout and public inspection terminology. This issue now tracks only the bounded MPMC broadcast variant.

## Goal

Add a bounded MPMC broadcast channel in which every active subscription observes every accepted publication in one committed order and the slowest active subscription applies backpressure.

## Contracts

- `bounded(capacity)` enforces the requested logical capacity and panics if `capacity` is zero.
- An asynchronous send waits when the slowest active subscription prevents reclamation; `try_send` reports `Full` without accepting the value.
- Every accepted publication remains available to every subscription that was active when it was accepted.
- New subscriptions start at the committed tail.
- Multiple producers expose one contiguous committed publication order; later claims cannot become visible before earlier claims.
- Sliding, drop-oldest, overwrite, and lag semantics remain out of scope.

## Acceptance criteria

- Cover fanout, concurrent producers, strict capacity, slow-subscription backpressure, reclamation, sender cancellation, receiver cancellation, disconnection, and panic-safe payload cloning.
- Benchmark 1 and 8 producers across representative fanout counts against semantically comparable lossless bounded broadcast channels.
- Compare shared paths with the existing unbounded implementation without forcing both retention policies through an abstraction that obscures either contract.
- Keep ring, cursor, retention, and sequencing types private; put cross-family machinery under `crate::internal` only when another implementation actually shares it.

The benchmark baseline in #208 is complete. Prior work and discussion: #117, #146, and #167.

onenewcode:
## Summary

Closes #213.

- Add `broadcast::mpmc::bounded`, a lossless broadcast channel whose shared backlog applies backpressure at the requested capacity.
- Share retention, subscription cursors, and receive registration with the unbounded channel.
- Preserve stack-based semaphore wake batches and complete every batch after a callback panics through one shared wake iterator.
- Compare capacities 1, 2, 8, 64, and 1024 on native threads and four Tokio workers against `async-broadcast`; measure bulk permit release and subscriber reclamation separately from setup and cleanup.

Validation: `cargo +1.86.0 x test` (586 passing tests), `cargo x check`, `cargo x lint`, `cargo x bench --no-run`, all 100 bounded ecosystem benchmark smoke cases, and six targeted Miri tests covering semaphore batches, panic/overflow recovery, and payload destruction outside locks.

## Design Notes

Each accepted publication has one commit order and stays readable by subscriptions active at commit time until they advance or are dropped. New subscriptions start at the committed tail. With no subscriptions, sends succeed and discard the value. Zero capacity is rejected.

Capacity counts unread messages in the shared backlog, not application work or pending sends. Receiving advances the subscription before returning the value; there is no acknowledgement or processing-completion barrier. Senders use semaphore notifications as retry hints, so capacity is not reserved and publication is not FIFO among waiting producers.

Payload cloning and destruction run outside the channel lock. Reclaimed capacity is released before touching payloads, and cancelled send registrations are released before their payloads are destroyed.


onenewcode:
### Throughput against `async-broadcast`, and where the gap comes from

Recording the numbers behind the benchmark section, since the ecosystem comparison is an acceptance criterion and the result is not uniformly favourable.

Local run on an aarch64 macOS laptop, `divan` medians over 10 samples of 4096 messages. The machine is noisy — treat any single figure as ±15% — so every pair below comes from the same run, and I have only paired rows that were measured together.

| shape | `async-broadcast` | `asyncband` | |
| --- | --- | --- | --- |
| cap 1024, 1 producer, 1 receiver | ~379–386 µs | ~339–388 µs | parity |
| cap 64, 1 producer, 1 receiver | 462 µs | 504 µs | 1.1× |
| cap 64, 1 producer, 8 receivers | 9.34 ms | 9.19 ms | parity |
| cap 64, 8 producers, 8 receivers | 10.4 ms | 12.0 ms | 1.15× |
| cap 1024, 1 producer, 32 receivers | 13.2 ms | 39.6 ms | 3.0× |
| cap 64, 1 producer, 32 receivers | 36.4 ms | 134 ms | 3.7× |
| cap 1024, 8 producers, 1 receiver | ~870–969 µs | ~3.05–3.33 ms | 3.5× |
| cap 64, 8 producers, 1 receiver | ~1.26–1.50 ms | ~4.85–5.03 ms | 3.6× |
| cap 1024, 8 producers, 8 receivers | 2.71 ms | 8.09 ms | 3.0× |
| cap 1024, 8 producers, 32 receivers | 7.93 ms | 21.3 ms | 2.7× |

Single-threaded `try_round_trip` is 24 ns against 40 ns.

**This is not specific to the bounded channel.** Running the existing unbounded comparison on the same machine shows the same gap in the same direction — `concurrent_producers` at 1/2/4/8 producers is 104/179/232/306 µs for `async-broadcast` against 166/303/334/436 µs for `asyncband` (1.4–1.7×), with `asyncband` ahead of `tokio::sync::broadcast` at 4 and 8 producers. The cost lives in the retention machinery that #117 already shipped and that #213 asks this PR to reuse, not in anything the bounded path adds on top.

Two properties account for it, and both are load-bearing rather than incidental:

1. **One `Arc` allocation per message.** Payloads are boxed so the receive path can move them out of the critical section, which is what keeps `T::clone` and `T::drop` — arbitrary user code that may reenter the channel — off the channel lock entirely. `async-broadcast` stores `T` inline and clones per receiver under its lock, which is cheaper for a `usize` and makes no such guarantee. `broadcast_mpmc_bounded_test.rs::bounded_message_destructors_run_outside_the_channel_lock` and `waitset_reentrancy_test.rs` pin this down; it is also what lets a single-receiver channel move the payload out instead of cloning it.
2. **One lock over backlog, cursors, and wait set.** That is what makes publish-and-drain a single critical section, which is in turn what makes the park path race-free — a receiver that finds nothing and registers still holds the lock, so a concurrent send cannot slip between the two steps and skip the wake-up.

**I tried the obvious fix and it did not survive measurement.** The one bounded-specific suspicion worth testing was that `try_send` holds the channel lock across `Arc::new`, which would serialize eight producers behind one allocator round trip. Hoisting the allocation above the lock and recovering the value with `Arc::into_inner` on rejection moved the contended shapes by ~3% — inside the noise — and regressed `cap 1024, 1 producer, 1 receiver` from ~340–390 µs to ~475 µs across both repetitions, because once sends actually block, every rejected `try_send` pays a malloc/free pair before `send` boxes the value for real. Reverted; the comment on `try_send` now records the measurement so the next person does not repeat it.

Closing the remaining gap means revisiting the `Arc` or sharding the lock, both of which change the unbounded channel's shipped guarantees as well. That seemed clearly out of scope for the PR that #213 describes, so I have left it alone — happy to open a follow-up issue for the fan-out and multi-producer shapes if you would like it tracked.

tests-integration/tests/broadcast_mpmc_bounded_test.rs
```diff
@@ -0,0 +1,810 @@
...
+        }
+        drop(rx2);
+        drop(rx1);
+
+        finished_tx.send(()).unwrap();
+    });
+
+    finished_rx
+        .recv_timeout(Duration::from_secs(10))
+        .expect("reclaimed message destructors must not run while the channel is locked");
```
ariesdevil: why not use [`assert_completes_without_deadlock`](https://github.com/apache/asyncband/blob/b68f3a20980af89f1258e4fc6eca21a44810f43a/tests-integration/src/lib.rs#L125) here?

ariesdevil (approved):
What can I say? Perfect.

PR #263, opened 2026-09-01
ci: link discussions to mailing list threads

Xuanwo:
## Summary

Configure stable subjects for GitHub Discussion notifications and add a workflow that resolves each mirrored `dev@asyncband.apache.org` thread through `asfml` and comments the archive URL back on the Discussion. The workflow also supports manual backfill for existing Discussions.

### DIFF
--- a/.github/workflows/discussion-thread-link.yml
+++ b/.github/workflows/discussion-thread-link.yml
@@ -1,0 +1,232 @@
+# Licensed to the Apache Software Foundation (ASF) under one
+# or more contributor license agreements.  See the NOTICE file
+# distributed with this work for additional information
+# regarding copyright ownership.  The ASF licenses this file
+# to you under the Apache License, Version 2.0 (the
+# "License"); you may not use this file except in compliance
+# with the License.  You may obtain a copy of the License at
+#
+#   http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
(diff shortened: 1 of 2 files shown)


.github/workflows/discussion-thread-link.yml
```diff
@@ -0,0 +1,232 @@
...
+                'json',
+              ]);
+              const emails = JSON.parse(stdout);
+              if (!Array.isArray(emails) || !emails.length) {
+                return null;
+              }
+
+              const candidate =
+                emails.find(email => normalizeSubject(email.subject) === normalizedTitle) ||
+                emails[0];
```
BewareMyPower, line 150: Falling back to the 1st email in the result seems not make sense. When could it happen?

BewareMyPower: It seems that it happened only when an incorrect discussion title was added. In this case, I tend to just fail instead of returning the 1st search result.

.github/workflows/discussion-thread-link.yml
```diff
@@ -0,0 +1,232 @@
...
+
+            function normalizeSubject(value) {
+              return (value || '')
+                .trim()
+                .replace(/^(re:\s*)+/i, '')
+                .toLowerCase();
+            }
+
+            async function runAsfml(args) {
+              const { stdout } = await execFileAsync('npx', ['--yes', 'asfml', ...args], {
```
BewareMyPower, line 119: It would be better to pin `asfml`'s version in case of security attack, though I confirmed it's maintained by yourself :)

```suggestion
              const { stdout } = await execFileAsync('npx', ['--yes', 'asfml@0.1.1', ...args], {
```

tisonkun (approved):
Let's use it as is and see.

Issue #216, opened 2026-08-26
build(channel): consolidate feature flags and documentation


tisonkun:
Parent: #206

## Resolution

The 0.7 feature and documentation structure is finalized:

- Keep `oneshot`, `mpsc`, `broadcast`, and `watch` as independent opt-in Cargo features matching the public delivery families that are implemented.
- Do not add a `channel` umbrella feature. It would not provide a stable audit, dependency, or compilation boundary beyond enabling unrelated APIs together.
- Do not create feature flags for producer topology or bounded versus unbounded policy. Those choices remain visible in module paths and constructors, such as `broadcast::mpmc::unbounded` and `mpsc::bounded`.
- Keep public modules and source families at direct crate-root paths. Shared implementation machinery remains private under `crate::internal` when it is genuinely shared.
- Document each feature and public module in the README and crate-level API map.

## Completed by

- #226 established the feature decision and documented the feature boundaries.
- #234 aligned the physical source tree with the public module families without changing feature names or public import paths.
- #235 is a release-documentation follow-up that makes the shipped channel topology more explicit; it does not change the decision above.

Future aggregate or topology-specific features require a concrete compile-time, dependency, binary-size, or audit boundary and can be proposed independently.

tisonkun:
Closing as completed. The final 0.7 decision is to retain independent `oneshot`, `mpsc`, `broadcast`, and `watch` features, with no `channel` umbrella and no topology- or capacity-specific feature flags. The issue body now records the outcome and the implementing PRs.

PR #265, opened 2026-09-01
feat(mpmc): add competing queues

### LINKED ISSUE #211: feat(mpmc): add a competing queue
Parent: #206

## Goal

Add bounded and unbounded MPMC queues in which cloneable receivers compete for each value and every accepted value is delivered exactly once.

## Direction

- Keep the public path at `asyncband::mpmc`.
- Use targeted waiter notification with cancellation handoff rather than waking every sender or receiver.
- Keep bounded capacity strict and unbounded sends synchronous.
- Keep implementation policy and storage types private.

## Acceptance criteria

- Cover clone counts, ordering, disconnection, draining, cancellation, exact-once delivery, and bounded progress under contention.
- Benchmark 1P/1C, 1P/8C, 8P/1C, and 8P/8C for bounded and unbounded modes.
- Avoid sustained order-of-magnitude regressions against semantically comparable MPMC channels.

Depends on #208. The implementation may share narrowly scoped machinery established by #209 without forcing single- and competing-consumer paths into one generic policy framework.

mxsm:
## Summary

- Add `asyncband::mpmc` bounded and unbounded queues with cloneable senders and receivers.
- Deliver every accepted value to exactly one competing receiver while receivers remain, preserve FIFO queue order, enforce strict bounded capacity, and drain buffered values before reporting sender disconnection.
- Use targeted FIFO waiter notification for ordinary sends and receives, pass selected notifications to the next waiter after cancellation, and reserve wake-all behavior for terminal disconnection.
- Add behavior, contention, trait, and cancellation tests plus bounded and unbounded ecosystem benchmarks for 1P/1C, 1P/8C, 8P/1C, and 8P/8C topologies.

Closes #211

## Design Notes

Queue storage, endpoint counts, capacity checks, and disconnection state are linearized under one private `Mutex<VecDeque<T>>`. Separate internal semaphores track blocked receivers and capacity-waiting bounded senders, so an ordinary queue transition selects one waiter without coupling public endpoints to storage or waiter types.

Dropping the final sender wakes all receivers, which first drain buffered values and then observe `Disconnected`. Dropping the final receiver releases buffered values and wakes every blocked bounded sender so each pending send returns its own unsent value.

The new `mpmc` Cargo feature is an independent compile, audit, and code-size boundary for the new public module. This keeps the implementation opt-in without placing MPMC APIs behind the existing `mpsc` feature; maintainers should confirm this boundary against the feature policy discussed in #216.

## Testing

- `cargo x check` — passed the no-feature build, the standalone `mpmc` and `event` feature builds, every other standalone feature build, and the all-feature build.
- `cargo x test --no-capture` — passed the full workspace suite, including 14 MPMC integration tests, upstream panic-recovery tests, and all documentation tests.
- `cargo x lint` — passed Clippy with warnings denied, rustfmt, Taplo, typos, license checks, and documentation generation.
- The MPMC tests cover clone counts, FIFO ordering, strict capacity, disconnection, buffered draining, unsent-value recovery, targeted wakeups, sender and receiver cancellation handoff, `T: Send + !Sync` endpoint traits, future `Send` traits, exact-once delivery, and bounded progress under 8P/8C contention.

## Performance

Command: `cargo bench -p benchmarks --bench ecosystem -- mpmc`

The following values are median time per 16,384-value batch from the optimized Windows run after merging the current `main`. Lower is better; the ratio compares Asyncband with the fastest result in the same row.

### Bounded, capacity 64

| Topology | Asyncband | async-channel | flume | Asyncband / fastest |
| --- | ---: | ---: | ---: | ---: |
| 1P/1C | 6.151 ms | 3.390 ms | 4.415 ms | 1.81x |
| 1P/8C | 24.52 ms | 34.51 ms | 26.57 ms | 1.00x |
| 8P/1C | 24.12 ms | 10.81 ms | 21.30 ms | 2.23x |
| 8P/8C | 7.219 ms | 7.681 ms | 7.045 ms | 1.02x |

### Unbounded

| Topology | Asyncband | async-channel | flume | Asyncband / fastest |
| --- | ---: | ---: | ---: | ---: |
| 1P/1C | 5.216 ms | 2.052 ms | 2.251 ms | 2.54x |
| 1P/8C | 24.04 ms | 3.661 ms | 10.69 ms | 6.57x |
| 8P/1C | 3.322 ms | 3.674 ms | 2.060 ms | 1.61x |
| 8P/8C | 5.897 ms | 4.315 ms | 3.513 ms | 1.68x |

The largest observed median gap was 6.57x for unbounded 1P/8C, below the order-of-magnitude regression threshold in #211. The private mutex-backed queue is expected to serialize more heavily than async-channel in some producer or consumer skews; the benchmark matrix is retained so future changes can track and improve those paths.


tisonkun (approved):
Thanks for your contribution @mxsm!

I think the API and implementation are generally good.

We may add `reserve` as MPSC for MPMC bounded as well later. And some performance improvement ideas can be landed later.

So far, I'm focusing on the correctness and API design and it looks good.

PR #286, opened 2026-09-05
test(pool): cover cancellation during bounded acquisition

QwQBiG:
## Summary

Motivated by reviewing OpenDAL's proposed Asyncband migration in apache/opendal#8219. Its [Memcached connection acquisition](https://github.com/apache/opendal/blob/a1387df5eea87ac446b5528381b113bbbc5053d7/core/services/memcached/src/core.rs#L191-L203) races `pool.get()` against a ten-second timeout, so an unfinished acquisition can be cancelled.

Add bounded pool tests for cancellation:

- while waiting for capacity;
- after notification but before the acquisition is polled again;
- during object creation, with another acquisition waiting.

These complement the existing recycling and replenishment cancellation tests by checking acquisition cancellation through the public pool API. They verify waiter progress, capacity recovery, and object reuse.

The tests use controlled polling without network calls or wall-clock timeouts. They check wake-driven progress without requiring FIFO notification order. Production code is unchanged; this is coverage for the existing cancellation contract, not a fix for an observed downstream failure.

## Validation

- 23 pool integration tests passed on stable and Rust 1.86.
- `cargo x test`: 476 tests and doctests passed.
- Targeted Clippy, rustfmt, typos, and Hawkeye checks passed.

Validation was performed locally in my WSL Linux.

### DIFF
--- a/tests-integration/tests/pool_acquire_cancelled_test.rs
+++ b/tests-integration/tests/pool_acquire_cancelled_test.rs
@@ -1,0 +1,208 @@
+// Licensed to the Apache Software Foundation (ASF) under one
+// or more contributor license agreements.  See the NOTICE file
+// distributed with this work for additional information
+// regarding copyright ownership.  The ASF licenses this file
+// to you under the Apache License, Version 2.0 (the
+// "License"); you may not use this file except in compliance
+// with the License.  You may obtain a copy of the License at
+//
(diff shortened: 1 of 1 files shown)


tisonkun (approved):
Thanks for your contribution @QwQBiG! LGTM.

Merging ...

PR #288, opened 2026-09-06
perf(completion): make terminal waits lock-free

### LINKED ISSUE #246: Explore a simpler waiter registration path for Completion
## Background

PR #245 makes `WaitSet` accept owned wakers so cloning, dropping, and waking user-provided wakers stay outside primitive state locks. Its callers now fall into three workload-driven polling shapes:

- monotonic state with a cheap lock-free probe can probe, clone, lock, and recheck;
- park-once primitives can clone before taking their state lock;
- primitives with a hot ready or buffered path can check under the lock, clone outside it only when needed, then lock and recheck.

`Completion` currently uses the third shape even though its `Pending` to `Completed` or `Abandoned` transition is monotonic. That preserves a cheap ready path, but makes its pending registration loop harder to understand.

## Goal

Evaluate whether `Completion` can use a simpler monotonic-state design, for example an atomic status alongside the existing `OnceLock<T>` and mutex-protected `WaitSet`. A successful design should retain a lock-free completed or abandoned probe and need only one waiter-lock acquisition on the common pending path, without executing `RawWaker` clone, drop, or wake callbacks under that lock.

## Requirements

- preserve the current public API, abandonment semantics, cancellation safety, and borrowed result;
- give a complete Acquire/Release argument for publishing the `OnceLock` value and abandonment;
- preserve safety under reentrant or panicking `RawWaker` clone, drop, and wake callbacks;
- compare the candidate with `main` using ready wait, pending registration and cancellation, single notification, and fan-out benchmarks;
- prefer the current implementation if a candidate does not lower the overall reasoning cost without a meaningful performance regression.

This should not introduce a generic polling-policy abstraction shared by unrelated primitives merely to hide their different load profiles.

ryux1:
Closes #246.

## Summary

- Make completed and abandoned `Completion` waits lock-free with one `OnceLock<Option<T>>`; pending registration remains cancellable.
- Reuse `WakerSet::register`, consolidate overlapping lifecycle tests, and cover terminal reads, pending repolls/cancellation, and notification fan-out in benchmarks.

## Design Notes

An uninitialized cell means pending; a stored `Some(value)` means completed and a stored `None` means abandoned. The single completer detaches the waiter set before initializing the cell, so observing a terminal result also observes waiter detachment. Pending polls and cancellation recheck the cell under the waiter mutex. Waker cloning follows the existing short-lock registration policy; replaced/cancelled wakers are dropped and notifications run after unlocking.

Local measurements compare main (`f056962`) with this branch on Apple M4 Max, macOS 26.6.2, Rust 1.98.0. Both use the same benchmark harness, excluding observer-container construction and handle cloning from fan-out. Values are medians of six per-run medians in nanoseconds, with run order balanced across main, an intermediate AtomicBool variant, and this implementation.

| Benchmark | main | This branch | Change |
| --- | ---: | ---: | ---: |
| `ready_wait` | 6.117 | 3.107 | -49.2% |
| `abandoned_wait` | 6.139 | 2.738 | -55.4% |
| `cancel_pending` | 11.700 | 12.150 | +3.8% |
| `repoll_pending` | 5.280 | 5.492 | +4.0% |
| `complete_then_wait` | 65.965 | 62.005 | -6.0% |
| `notify_pending` | 94.505 | 90.465 | -4.3% |
| `fanout/2` | 174.150 | 164.350 | -5.6% |
| `fanout/8` | 420.250 | 387.500 | -7.8% |
| `fanout/32` | 1292.000 | 1164.000 | -9.9% |

Terminal reads improve substantially; pending cancellation and repolling pay roughly 0.45 ns and 0.21 ns respectively. Small fan-out differences are noisy. Completion/notification throughput is comparable to the AtomicBool variant, without maintaining a separate terminal flag.

Validation: `cargo +stable x test` and `cargo +1.86.0 x test` (484 tests each), `cargo x check`, `cargo x lint`, `cargo +stable x bench --no-run`, and all 14 Completion integration tests under Miri with strict provenance, symbolic alignment, and seeds `0..4`.


tisonkun (commented):
Generally LGTM.

I'm considering two minor improvements. Would test out tomorrow and get this merged.

tisonkun (approved):

Issue #272, opened 2026-09-03
Explore a higher-level task group primitive


tisonkun:
## Context

`WaitGroup` is a low-level lifetime barrier: cloned RAII handles represent outstanding work, and waiters observe when all handles are gone. It deliberately does not own tasks, collect results, propagate errors, or define cancellation and parent-child relationships.

Some use cases need a higher-level abstraction for a group of related asynchronous tasks. That should be designed as a separate primitive rather than changing `WaitGroup`. `ForkJoin` was an initial placeholder, not a proposed public name.

## Ecosystem survey

- Rust's [`tokio::task::JoinSet<T>`](https://docs.rs/tokio/latest/tokio/task/struct.JoinSet.html) owns runtime-spawned tasks, yields homogeneous results in completion order, and aborts remaining tasks when dropped. It is primarily controlled by one mutable collection owner.
- Rust's [`tokio_util::task::TaskTracker`](https://docs.rs/tokio-util/latest/tokio_util/task/task_tracker/struct.TaskTracker.html) is cloneable and supports registration without mutable access. Waiting requires the tracker to be both closed and empty; completed tasks release their storage immediately, results are not retained, and dropping the tracker does not abort tasks.
- Go's [`errgroup.Group`](https://pkg.go.dev/golang.org/x/sync/errgroup) groups goroutines belonging to one operation, waits for them, returns the first non-nil error, can cancel a derived context on failure, and optionally limits concurrency.
- Java's [`StructuredTaskScope`](https://docs.oracle.com/en/java/javase/26/docs/api/java.base/java/util/concurrent/StructuredTaskScope.html) makes one owner responsible for a lexical task scope. It forks subtasks, joins them under an explicit policy, prevents new forks after joining, and closes by cancelling and waiting for unfinished subtasks.
- Swift's [`TaskGroup`](https://developer.apple.com/documentation/swift/taskgroup) is a non-escaping structured scope for dynamically created child tasks. Children inherit parent context, results form an asynchronous sequence, and the scope cannot return while children remain.
- Kotlin's [`coroutineScope`](https://kotlinlang.org/api/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines/coroutine-scope.html) establishes a parent-child task tree: the parent waits for children, while failure and cancellation propagate through the scope.

[...]

PR #308, opened 2026-09-13
feat: add task group

Colerar:
## Summary

This PR implements the first `TaskGroup` proposed in the issue discussion.

Issue: #272

The design rationale and alternatives are discussed in [this issue comment](https://github.com/apache/asyncband/issues/272#issuecomment-5636767531).

## Main API

```rust
let (mut group, registrar) = TaskGroup::new();

let first = tokio::spawn(registrar.track(worker_a()).unwrap());
let second = tokio::spawn(registrar.track(worker_b()).unwrap());

group.close();

while let Some(output) = group.join_next().await {
    process(output);
}

first.await.unwrap();
second.await.unwrap();
```

## Scope of this PR

This version intentionally does not provide:

- cancellation or a cancellation token
- forced interruption or task abortion
- `try_join()`, `race()`, or `race_ok()` result policies
- bounded concurrency or output backpressure
- multiple result consumers


tisonkun (changes requested):
This doesn't seem like a common primitive. When would you need such a primitive?

I may consider this `join`-alike primitive a subset of https://docs.rs/futures-concurrency.

While I may consider to include some functions of `futures-concurrency` in asyncband, it doesn't look like in this form:

1. Why close?
2. Why all tasks must output the same type `T`?

I may expect some real world use cases for motivation and would consider if we can have a more generic/unified form as `futures-concurrency` has already designed.

PR #313, opened 2026-09-14
feat(watch): allow non-Clone channel values

### LINKED ISSUE #311: Relax the Clone bound on watch channel construction
## Motivation

`watch::channel` currently requires `T: Clone`, although constructing the channel and publishing values do not clone the payload. This prevents callers from using non-Clone state with the existing publication, replacement, and change-notification APIs.

Only the owning read methods, `Receiver::get` and `Receiver::recv`, need `T: Clone`. Their method-level bounds already express that requirement.

## Proposed change

Relax the constructor from:

```rust
pub fn channel<T: Clone>(initial: T) -> (Sender<T>, Receiver<T>)
```

to:

```rust
pub fn channel<T>(initial: T) -> (Sender<T>, Receiver<T>)
```

Keep the existing `T: Clone` bounds and return types on `Receiver::get` and `Receiver::recv`. Construction, `send`, `send_replace`, subscription, handle cloning, `has_changed`, and `changed` should be usable with a non-Clone payload.

This is a compatible relaxation: existing callers retain the same types and behavior. It does not introduce borrowed reads or change the channel to store or return `Arc<T>`; callers can already select `T = Arc<State>` when they want shared owning snapshots.

## Implementation and validation

- Update the [constructor](https://github.com/apache/asyncband/blob/8d6db15d7b1f9c9b0631c347ed7492587654e3b3/asyncband/src/watch/mod.rs#L93) and document that only owning reads require Clone.
- Add focused coverage using a payload that does not implement Clone. Exercise publishing and observing a change, and verify that `send_replace` returns ownership of the previous value. Existing [Watch integration tests](https://github.com/apache/asyncband/blob/8d6db15d7b1f9c9b0631c347ed7492587654e3b3/tests-integration/tests/watch_test.rs) provide the test style and cover existing Clone-based behavior.
- Use deterministic coordination for any pending-wait coverage, and run the applicable `cargo x` checks.
- Record the newly supported use with non-Clone values in the changelog.

The constructor is in [`asyncband/src/watch/mod.rs`](https://github.com/apache/asyncband/blob/8d6db15d7b1f9c9b0631c347ed7492587654e3b3/asyncband/src/watch/mod.rs); the owning-read bounds are already localized on [`get` and `recv`](https://github.com/apache/asyncband/blob/8d6db15d7b1f9c9b0631c347ed7492587654e3b3/asyncband/src/watch/mod.rs#L268).

TayfurYldz:
## Summary

- relax `watch::channel` so construction, publication, replacement, subscription, and change tracking do not require `T: Clone`
- keep `Clone` localized to `Receiver::get` and `Receiver::recv`
- add focused integration coverage with a non-`Clone` payload, including `send_replace` ownership
- document the contract and changelog entry

Closes #311.

## Testing

Repository CI is the authoritative validation for `cargo x lint`, `cargo x check`, `cargo x bench --no-run`, `cargo x test --no-capture`, and Miri.


asyncband/src/watch/mod.rs
```diff
@@ -81,6 +81,8 @@ use crate::internal::wakerset::WakerToken;
 /// Creates a watch channel with an initial value.
 ///
 /// The receiver returned by this function considers the initial value already observed.
+/// The value type does not need to implement `Clone` for construction, publication, or change
+/// notification. Only owning reads through [`Receiver::get`] and [`Receiver::recv`] require it.
```
tisonkun:
```suggestion
```

Too many deatils.

tests-integration/tests/watch_non_clone_test.rs
tisonkun, line 1: Is this really needed?

tisonkun (approved):
Thanks for your contribution @TayfurYldz !

Merging...

Issue #303, opened 2026-09-11
Explore release automation with Apache Trusted Releases and weekly candidates


tisonkun:
## Motivation

Reduce the manual work in the Asyncband release lifecycle by exploring Apache Trusted Releases (ATR) integration and a weekly release attempt when new commits are available.

Asyncband already uses crates.io Trusted Publishing for approved final tags. This work concerns candidate preparation, source-artifact signing and staging, voting, and the handoff to that existing publication workflow.

OpenDAL provides useful references: [weekly candidate preparation](https://github.com/apache/opendal/blob/af11b6ee9ad1e4df1ff7b1e61156b1f3c4a74357/.github/workflows/weekly_release.yml), [ATR composition](https://github.com/apache/opendal/blob/af11b6ee9ad1e4df1ff7b1e61156b1f3c4a74357/.github/workflows/release-compose.yml), and [release lifecycle synchronization](https://github.com/apache/opendal/blob/af11b6ee9ad1e4df1ff7b1e61156b1f3c4a74357/.github/workflows/release_lifecycle.yml). Adapt the relevant parts to Asyncband's release process.

## Scope to explore

- Define a weekly source cutoff and comparison baseline. Skip runs without new commits and avoid duplicate candidates when a release is already being prepared or voted on. Decide how to handle new commits during an active vote.
- Prepare a reviewable version and changelog update, including pre-1.0 breaking changes, using the existing `cargo x` checks and semver workflow.
- Evaluate [ATR Trusted Publishing](https://releases.apache.org/docs/trusted-publishing): reproducible artifacts, project signing keys, and permitted compose/vote/finish workflows. Its ASF configuration is separate from the already configured crates.io OIDC publishing.
- Preserve the [Incubator's PPMC and IPMC approval process](https://incubator.apache.org/policy/incubation.html#releases). Verify ATR's two-round podling workflow and the remaining release-manager decisions; OpenDAL's PMC workflow is not a direct substitute.
- Track the source and release commits, version, RC tag, artifact digests, and ATR revision. Retries should resume the same candidate and retain the exact voted artifacts.
- Evaluate ATR staging and [approved source publication](https://releases.apache.org/docs/promoting-to-release), then connect successful IPMC approval to the final tag and protected crates.io publication. Handle partial successes before retrying.

## Proposed delivery

[...]

PR #318, opened 2026-09-15
ci: prepare reproducible source archives for automated signing

Xuanwo:
## Summary

Prepare Asyncband's source-signing workflow for an ASF Infra-managed `GPG_SECRET_KEY` request and ASF Security review, following the [automated release signing procedure](https://infra.apache.org/release-signing.html#automated-release-signing). Related to #303.

A single trusted job packages the committed source with `git archive --format=tar.gz -9`, writes its SHA-512 checksum with `shasum`, and optionally signs and verifies the archive with GPG. PRs and default manual runs produce unsigned artifacts. After provisioning, the release manager manually runs **Source candidate** on `apache/asyncband`'s `main` branch with **sign** enabled and downloads the `signed-source` GitHub Actions artifact. ATR upload is a follow-up integration; this workflow does not stage candidates, start votes, create tags, or publish releases.

The release instructions require independent byte-for-byte reproduction with Git and `cmp` on trusted hardware against the actual staged archive before publication. Before signing is enabled, ASF Security must approve the workflow, Infra must install `GPG_SECRET_KEY`, and the public key and `SOURCE_SIGNING_FINGERPRINT` must be configured. The signing command expects an empty passphrase on the Infra-managed key.

## Reproducibility evidence

Verified commit `6a6e8901215c2ea82326ed44104f3277ff832482`. The archive downloaded from [Source candidate run 34992337603](https://github.com/apache/asyncband/actions/runs/34992337603) on Ubuntu 24.04 (Git 2.55.0) matches the archive independently reproduced from a fresh checkout on macOS Apple Silicon (Apple Git 2.54.0). The documented reproduction commands passed against the downloaded CI archive, including `cmp`.

SHA-512 of `apache-asyncband-0.7.2-incubating-src.tar.gz`:

```text
2c4159a54ec50fa79ceb91299eff35e99ffcf9373986fa252451360193218e87288001e3fe43596c8d89f471bbc0788842b4b238f45e117aaf66ac2c1ee8ed22
```

The shell checks rejected tracked edits, output-directory reuse, and a modified downloaded archive. Production signing remains untested until Infra provisions the key. Each future candidate still requires independent reproduction against its actual staged bytes.


.github/workflows/source.yml
tisonkun, line 1: Since we have single target to sign, we don't need two stages (jobs) and do upload/download.

xtask/src/source.rs
tisonkun, line 1: I suppose these logics would be in `main.rs` and you may use some crates for gzip/tar/shasum instead of calling commands as utils may not avaiable or varies in some platforms.

.github/workflows/source.yml
Xuanwo: makes sense, removed.

xtask/src/source.rs
Xuanwo: Yep, merged.

.agents/skills/release/references/candidate.md
```diff
@@ -68,24 +68,26 @@ Wait for the `Release` GitHub Actions workflow to pass. The workflow validates t
...
 RC_TAG="v${VERSION}-rc.${RC}"
 SOURCE_DIR="apache-asyncband-${VERSION}-incubating-src"
 ARTIFACT_DIR="${RELEASE_DIR}/dist"
-mkdir -p "${ARTIFACT_DIR}"
 git verify-tag "${RC_TAG}"
-git archive --format=tar --prefix="${SOURCE_DIR}/" "${RC_TAG}" \
-  | gzip -n -9 > "${ARTIFACT_DIR}/${SOURCE_DIR}.tar.gz"
+test "$(git rev-parse HEAD)" = "$(git rev-parse "${RC_TAG}^{commit}")"
+cargo x --help
+cargo x source --help
```
tisonkun: This should not exist in the shell script? They seem to be information query, rather than the one that would be executed.

.agents/skills/release/references/candidate.md
```diff
@@ -68,24 +68,26 @@ Wait for the `Release` GitHub Actions workflow to pass. The workflow validates t
...
+cargo x source --output "${ARTIFACT_DIR}"
 (
   cd "${ARTIFACT_DIR}"
-  shasum -a 512 "${SOURCE_DIR}.tar.gz" > "${SOURCE_DIR}.tar.gz.sha512"
   gpg --armor --detach-sign --local-user "${ASF_GPG_FINGERPRINT}" \
     "${SOURCE_DIR}.tar.gz"
```
tisonkun, line 89: I suppose we should let `cargo x source` do the signing job as well? Or else we may not need a `cargo x source` command.

You may please read the output of agent.

.github/workflows/source.yml
```diff
@@ -0,0 +1,95 @@
...
+      - xtask/**
+      - Cargo.toml
+      - Cargo.lock
+      - .agents/skills/release/**
+  workflow_dispatch:
+    inputs:
+      sign:
+        description: Sign with the Infra-managed key (requires ASF Security approval)
+        type: boolean
+        default: false
```
tisonkun, line 32: Who would trigger the `workflow_dispatch` event?

.agents/skills/release/references/infrastructure.md
```diff
@@ -32,3 +32,36 @@ Trusted Publishing is already configured for `asyncband`: repository `apache/asy
...
+
+Once approval, provisioning, and public-key publication are complete, manually run **Source candidate** on `main` with **sign** enabled. Record the run URL, full source commit, checksum, and signing fingerprint. Download the `signed-source` artifact and retain the archive, checksum, and signature as the candidate bytes. A retry should reuse an existing signed bundle when available.
+
+### Reproduce before staging and voting
+
+The release manager must reproduce the source archive on trusted hardware outside GitHub Actions, using a fresh checkout of the exact commit printed in the workflow summary. Compare against the actual downloaded archive, not merely its checksum file. Use Git and the committed `Cargo.lock`. The command uses `git archive` for the committed tree, `flate2` with its Rust backend and fixed gzip header for compression, and `sha2` for the checksum. It records the commit and checksum; Rust compares the full archive bytes. Run the following from that clean checkout, after setting `DOWNLOADED_ARCHIVE` to the absolute path of the downloaded `.tar.gz`:
+
+```shell
+cargo x --help
+cargo x source --help
```
tisonkun: Ditto. `--help` should not be included in the exuecting script.

xtask/src/main.rs
```diff
@@ -63,10 +69,114 @@ enum SubCommand {
...
+        let revision = command_output(revision);
+        let revision = std::str::from_utf8(&revision)
+            .expect("invalid Git revision")
+            .trim();
+        println!("Commit: {revision}\nVersion: {version}");
+
+        // Git owns the committed file inventory, modes, symlinks, and timestamps.
+        let mut archive = find_command("git");
+        archive
+            .args(["-c", "tar.umask=0022", "archive", "--format=tar"])
```
tisonkun: As we use `git` already, having something like `git archive --format=tar.gz` would package `tar.gz` without introducing the large `flate2` dependency.

tisonkun (changes requested):
Let's just drop `cargo x source` and use shell commands.

I found Rust is very weak in this area.

.github/workflows/source.yml
tisonkun: I suppose we may leverage some ATR actions to just upload the package files there.

tisonkun:
Superseded by https://github.com/apache/asyncband/pull/319.

PR #319, opened 2026-09-15
ci: compose signed source releases from RC tags

tisonkun:
## Summary

Pushing an RC tag such as `v0.7.3-rc.1` builds its source archive, generates a SHA-512 checksum, signs it with the ASF project key, and uploads the three release files through [apache/tooling-actions/upload-to-atr](https://github.com/apache/tooling-actions/blob/fa721a0b176d713807b574da721b96545b587eea/upload-to-atr/README.md).

Pull requests changing `release-compose.yml` or `.asf.yaml` run packaging checks against the PR checkout and skip signing and uploading.

## Design Notes

- Follow the existing `vX.Y.Z-rc.N` tag convention. Read the release version from the tag, require it to match `asyncband/Cargo.toml`, and verify the checked-out tag commit belongs to main, as the existing crate release workflow does.
- Build on Ubuntu 24.04 with `git archive | gzip -n -9`, fixed archive permissions, and no gzip filename or timestamp. Archive the tag's exact commit and record the tag, commit, and checksum in the workflow summary.
- Split packaging from signing and uploading. `compose` has read-only repository permission and no OIDC permission. `sign-and-upload` runs only for upstream RC tag pushes, downloads the exact source artifact by ID, and uses the existing `release` environment for approval and signing secrets. Only this job has OIDC permission. The ASF-allowlisted `crazy-max/ghaction-import-gpg` action imports the key, caches its passphrase, and removes the key during post-job cleanup; signing verifies the expected primary fingerprint.
- Upload `.tar.gz`, `.tar.gz.asc`, and `.tar.gz.sha512` under ATR version `X.Y.Z`; each upload creates a new revision. Record that revision alongside the RC tag before voting. Rerunning a failed signing/upload job reuses the source artifact and signs it again, so inspect ATR's state before retrying an upload failure.
- `.asf.yaml` registers the repository and compose workflow path for tag-triggered uploads. Voting, final publication, and the existing crates.io workflow are unchanged.

For the [Trusted Publishing signing-key application](https://releases.apache.org/docs/trusted-publishing#how-to-set-up-trusted-publishing), submit this pipeline for ASF Security's reproducibility confirmation and request a key with a UID such as `Asyncband Automated Release Signing <private@asyncband.apache.org>`. Configure the following in the existing `release` environment of `apache/asyncband`:

- Environment secret `GPG_SECRET_KEY`: ASCII-armored private key provisioned by ASF Infrastructure.
- Environment variable `SOURCE_SIGNING_FINGERPRINT`: full uppercase primary-key fingerprint.
- Optional environment secret `GPG_PASSPHRASE`: passphrase if the supplied key is protected.

Publish the public key through the existing project `KEYS` management and confirm ATR imports it and applies the compose policy.

Validation: executed the packaging script against the existing annotated `v0.7.2-rc.1` tag in a disposable Linux environment and verified its archived source, version, and checksum. PR packaging, invalid RC tags, manifest mismatch, checkout/tag mismatch, and unmerged candidate rejection pass. The pinned GPG action was exercised with an encrypted temporary key: signing, detached verification, and post-job key cleanup pass, including fingerprint mismatch and corrupted-archive failure paths. `actionlint`, the current ASF allowlist, `cargo x lint`, and `git diff --check` pass. Actual ATR upload awaits the project signing key and live trust configuration.

Related to #303.


tisonkun:
INFRA ticket link - https://issues.apache.org/jira/browse/INFRA-28407

.github/workflows/release-compose.yml
```diff
@@ -0,0 +1,145 @@
...
+
+name: Compose source release
+
+on:
+  push:
+    tags: [ "v*.*.*-rc.*" ]
+  pull_request:
+    paths:
+      - .github/workflows/release-compose.yml
+      - .asf.yaml
```
ppkarwasz, line 26: Why is this also triggered by `pull_request`? Or is this temporary just to test this workflow?

.github/workflows/release-compose.yml
```diff
@@ -0,0 +1,145 @@
...
+            cat "${source}.tar.gz.sha512"
+          } >> "${GITHUB_STEP_SUMMARY}"
+      - id: sign
+        name: Sign with the ASF project key
+        if: ${{ github.event_name == 'push' && startsWith(github.ref, 'refs/tags/v') && github.repository == 'apache/asyncband' }}
+        env:
+          ARCHIVE: ${{ steps.source.outputs.archive }}
+          SIGNING_KEY: ${{ secrets.GPG_SECRET_KEY }}
+          SIGNING_PASSPHRASE: ${{ secrets.GPG_PASSPHRASE }}
+          SIGNING_FINGERPRINT: ${{ vars.SOURCE_SIGNING_FINGERPRINT }}
```
ppkarwasz: [`crazy-max/ghaction-import-gpg`](https://github.com/crazy-max/ghaction-import-gpg), which is allowlisted by the ASF, might be an easier way to ensure that the key isn't left on the (ephemeral) GitHub runner VM.

.github/workflows/release-compose.yml
```diff
@@ -0,0 +1,145 @@
...
+  LC_ALL: C
+
+concurrency:
+  group: source-release-${{ github.event_name == 'pull_request' && github.event.number || 'compose' }}
+  cancel-in-progress: ${{ github.event_name == 'pull_request' }}
+
+jobs:
+  compose:
+    name: Compose source release
+    runs-on: ubuntu-24.04
```
ppkarwasz, line 45: I would run it in the `release` environment, which forces a review from @jbonofre, @Xuanwo, @PragmaTwice or you to sign on the release:

```suggestion
    runs-on: ubuntu-24.04
    environment:
      name: release
      url: https://releases.apache.org/projects/asyncband
```

ppkarwasz (changes requested):
Hi @tisonkun,

This looks pretty solid to me, although I have two suggestions:

- You can use the `release` environment to require an additional manual review, before the workflow starts.
- You can split the workflow into a `compose` job and a `sign-and-upload` job. The first job would not require privileges. You can even generate a build attestation for the first job and upload it in the second.

If you use environments, you need to specify it in the INFRA ticket, so they put the secrets in the right environment.

.github/workflows/release-compose.yml
tisonkun: This tests the packaging steps work well and it would not upload anything.

.github/workflows/release-compose.yml
tisonkun: OK. Let me integrate it.

.github/workflows/release-compose.yml
tisonkun: Sounds good. Let me try to integrate this idea.

tisonkun:
> You can split the workflow into a compose job and a sign-and-upload job. The first job would not require privileges. You can even generate a build attestation for the first job and upload it in the second.

This is possible. I merge the previous three jobs into one in https://github.com/apache/asyncband/pull/319/changes/db215e72b6f085fb48083a554b953989167d50a7 because I'd prefer to avoid upload/download-artifacts back and forth. But limiting privileges can be a good argument for splitting jobs.

tisonkun:
@ppkarwasz Your comments are addressed. PTAL.

ppkarwasz (approved):
LGTM! :100:

Actually I think that the `git archive` with the right parameters, signing and uploading should be a common action in `tooling/actions` that many PMC can reuse.

